PR #2, opened 2021-02-09
Support for KeyShared subscriptions and fixes acknowledgement handling for shared subscriptions

csthomas1:
Adds "Key_Shared" as an available subscription type to ConsumePulsar and ConsumePulsarRecord.

Also fixes some issues with the use of acknowledgeCumulative* regardless of subscription type, and use of acknowledge*Async when the async flag is enabled.


david-streamlio:
LGTM

PR #3, opened 2021-02-10
Allow empty demarcator; insert demarcators only between messages

csthomas1:
Alters demarcator support in ConsumePulsar and ConsumePulsarRecord as follows:

- An empty demarcator value is now valid (property validation previously required a non-empty value).
- Demarcators will now be inserted only between two consecutive messages in the same batch. Previously, a demarcator was always appended to each batched message regardless of whether another message followed it, which resulted in each batch having a dangling demarcator.

### DIFF
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
@@ -24,6 +24,7 @@ import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
```diff
@@ -219,7 +219,7 @@ private void consumeMessages(ProcessContext context, ProcessSession session, fin
        }
 
        RecordSchema schema = null;
-       final BlockingQueue<Message<byte[]>> parseFailures = new LinkedBlockingQueue<Message<byte[]>>();
+       final Queue<Message<byte[]>> parseFailures = new LinkedList<Message<byte[]>>();
```
david-streamlio, line 222: Why do you no longer use a BlockingQueue?

nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
```diff
@@ -356,7 +356,7 @@ public Object call() throws Exception {
     	}
     }
     
-    private void handleFailures(ProcessSession session, BlockingQueue<Message<byte[]>> parseFailures, byte[] demarcator) {
+    private void handleFailures(ProcessSession session, Queue<Message<byte[]>> parseFailures, byte[] demarcator) {
```
david-streamlio, line 359: Why do you no longer use a BlockingQueue?

nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
csthomas1: Hi David -- it didn't seem to be providing any benefit, given the way the logic in consumeMessage() and handleFailures() was implemented. consumeMessages() appeared to instantiate the Queue as a local variable, and append to it in a synchronous manner. Similarly, handleFailures appeared to iterate over the queue's contents in a synchronous manner. The use of the BlockingQueue and calls to queue.poll() therefore seemed confusing and potentially unnecessary.

I've been wrong before though :-). Maybe I missed something?

nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
csthomas1: See above.

nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
david-streamlio: I don't recall exactly why I used it either, but I am of the "if it isn't broke, don't fix it" mentality.  Now, if the performance has improved and there are no lost messages, then I am ok with the change.

csthomas1: There should be some theoretical performance improvement using the LinkedList vs. the LinkedBlockingQueue (due to the absence of locking in a situation that I contend doesn't require it anyway), but in normal operation one probably isn't going to see sufficiently frequent parse failures for it to matter.

I just pushed up a version using the LinkedBlockingQueue in place of the LinkedList.

Issue #25, opened 2021-12-17
Pulsar consumer/producers leak with percentage of unacked messages when using async mode


nicknezis:
When running NIFI alongside a Kubernetes deployed Pulsar cluster, we find that Pulsar metrics in cluster show producers and consumers are stable for a while, but eventually start growing. After some time, the growth reaches a limit and messages stop flowing.

We are making connections from NIFI to the Pulsar brokers through the Pulsar Proxy. Also we have the NIFI PulsarConsumer and PulsarProducer configured to use Async communication.

When testing with a simple Java client, we find that we can read data out of Pulsar, and publish to NIFI (using HTTP Listener processor). Because of this, we feel we have isolated the issue to the NIFI processors and/or controller services.

We are also running a few different Pulsar client controller services. Not sure if that would impact the performance. Essentially we have one for publishing and another for the consumption of data.

Even before we see Pulsar producers and consumer counts climbing, we do see NIFI log messages that consistently imply a smaller ack count (compared to the received message count). While the consumer processor is running, we will see old data retransmitted. We fear this failure to ack all received messages may be contributing to the eventual failure of the flow of data.

david-streamlio:
Resolved

PR #29, opened 2021-12-17
Bugfix/message loop consumer receive

csthomas1:
Includes two bug fixes, one major and one minor:
- In loop conditions of the message receive loops in ConsumePulsar, ConsumePulsarRecord, and AbstractPulsarConsumerProcessor, moved the consumer.receive() call to occur AFTER the comparison of the current batch size against the maximum batch size. The previous approach (consumer.receive() BEFORE the batch size check) was causing an extra message to be received and go unhandled whenever messages were still available and the batch size limit was reached. This resulted in unnecessary message timeouts/redeliveries, and became especially problematic when the individual message was part of a large internal batch, which tended to happen in high volume or "bursty" dataflows.
- Tweaked the acknowledgeAsync() executor service task in ConsumePulsar and ConsumePulsarRecord to make a call to Future.get() for consistency with the acknowledgeCumulativeAsync() executor service tasks.


nicknezis:
Fixes #25

PR #91, opened 2025-09-07
Fixed Issue 85

david-streamlio:
Added support for OAuth2 authentication using clientId, clientSecret instead of Private key file


david-streamlio:
Tested manually and confirmed against a StreamNative cluster.

PR #157, opened 2026-08-28
fix: pool and close Pulsar producers instead of leaking every one

### LINKED ISSUE #156: PublisherPool never pools and never closes: every Pulsar producer is leaked
Found while reviewing the publish path, which has no real-broker coverage at all.

`PublisherPool.publisherQueue` is **never written to** — nothing in the class calls `offer()` or `add()`, only `remove()` and `poll()`. Three consequences follow:

1. `obtainPublisher()` goes straight to `createLease()` on every call, building a brand new Pulsar producer even when an identical idle one exists. The class is a factory with a dead queue attached.
2. A lease’s `close()` takes the pool-is-open branch and calls `publisherQueue.remove(this)` on a permanently empty queue. `super.close()` — which flushes and closes the producer — **never runs**.
3. `PublisherPool.close()` drains that same empty queue, so it closes nothing.

**Every producer this bundle opens, and its broker connection, is leaked.** Proven with a probe against `main`:

```
verify(producer, never()).close();   // after lease.close()  -> PASSED
verify(producer, never()).close();   // after pool.close()   -> PASSED
// 5 producers created, 0 closed
```

`PublishPulsarRecord` compounds it: it obtains a lease inside the per-FlowFile loop (`PublishPulsarRecord.java:119`) and never closes it, so a busy flow opens one producer per FlowFile. `PublishPulsar` at least caches a lease per topic, but its `close()` hits the same no-op path.

`AbstractPulsarProducerProcessor` also has no `@OnStopped`, so the pool built in `@OnScheduled` is simply abandoned on stop. The asymmetry with the consumer side is the tell — consumers are closed correctly through the LRU cache; producers are not closed at all.

**Impact:** broker-side producer and connection exhaustion under sustained publishing, with no error surfaced until the broker starts refusing connections.

david-streamlio:
Closes #156.

**Change**
- The idle-lease queue is now **per topic** — a lease wraps a producer bound to one topic, so a single shared queue could hand a caller a producer for the wrong topic. Releasing a lease returns it for reuse; closing the pool closes every producer it holds.
- `PublisherLease.reset()` clears the cumulative message count when a lease is handed back out. Without this, reuse would make `msg.count` over-report on the second FlowFile and every one after — the trap flagged in the review.
- `PublishPulsarRecord` releases its lease in a `finally`, after reading `complete()`.
- `AbstractPulsarProducerProcessor` gains `@OnStopped` to close the pool, and `@OnScheduled` now closes any pool left from a previous scheduling rather than overwriting the field and stranding it.

**That last point was found by the test, not by reading.** `stoppingTheProcessorClosesItsProducers` failed on the first run because scheduling twice replaced the pool without closing the first one.

**Verification**
- `PublisherPoolLifecycleTest` (unit): reuse, per-topic isolation, close propagation, counter reset. **Three of five fail on `main`.**
- `PublishPulsarLifecycleIT` (real broker): asserts against the broker's own stats endpoint, which is the only place this leak is visible. 25 FlowFiles now hold **one** producer instead of 25, and stopping or restarting leaves **zero** connected.

Full build green: 226 unit + 8 integration.

### DIFF
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarProducerProcessor.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarProducerProcessor.java
@@ -23,6 +23,7 @@ import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.nifi.annotation.lifecycle.OnScheduled;
+import org.apache.nifi.annotation.lifecycle.OnStopped;
 import org.apache.nifi.annotation.lifecycle.OnUnscheduled;
 import org.apache.nifi.components.AllowableValue;
 import org.apache.nifi.components.PropertyDescriptor;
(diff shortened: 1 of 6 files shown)


david-streamlio:
Superseded by #158, which landed the stronger implementation of the same fix.

Two things #158 does that this PR did not:

- it tracks **in-use** leases as well as idle ones, so closing the pool closes every producer it created rather than only those sitting in the idle queue;
- it handles the race of a producer being created while the pool is closing.

Two pieces from here are still worth having and are not in #158, so they are going up as a follow-up:

- **the `@OnScheduled` guard** — `init()` still overwrites the pool field without closing the previous pool, so scheduling twice without an intervening stop strands the first pool's producers. This was found by an integration test failing, not by reading;
- **`PublishPulsarLifecycleIT`** — real-broker verification against the broker's own stats endpoint, which is the only place this leak is actually visible. #158 was unit-verified only (no Docker on that machine).

Closing in favour of #158; no work is lost.

PR #168, opened 2026-08-28
fix: acknowledge Pulsar messages only after the session carrying them commits

### LINKED ISSUE #167: ConsumePulsar acknowledges messages it then discards, losing them on any write error
Found while writing the restart tests for #18. This is a much more plausible cause of "messages disappear" than anything those tests turned up.

`ConsumePulsar` acknowledges messages to the broker either **before** their content is written, or **after** the session that would have carried them has been rolled back. An acknowledged message is gone from Pulsar: it is not redelivered, and it never reached a FlowFile. There is no error relationship and nothing is routed anywhere.

Three variants, on current `main`:

### 1. Synchronous, non-shared subscription — rollback then acknowledge

`ConsumePulsar.java:308`

```java
} catch (final IOException ioEx) {
    getLogger().error("Unable to create flow file ", ioEx);
    session.rollback();
    if (!shared) {
        consumer.acknowledgeCumulative(lastMsg);   // <-- everything up to lastMsg is now gone
    }
    return;
}
```

`session.rollback()` discards the FlowFiles, and the very next statement tells the broker that every message up to and including `lastMsg` has been handled. This is the clearest case: the two lines directly contradict each other.

### 2. Synchronous, Shared subscription — acknowledge before write

`ConsumePulsar.java:292`

```java
if (shared) {
    consumer.acknowledge(msg);     // acked here...
}

byte[] data = msg.getData();
if (data != null && data.length > 0) {
    out.write(data);               // ...written here
}
```

Each message is acknowledged before its content is written. If the write throws, the `catch` rolls the session back and those messages are already gone.

### 3. Asynchronous, Shared subscription — same ordering

`ConsumePulsar.java:163` submits `acknowledgeAsync(msg)` before the write block at `ConsumePulsar.java:174`. Same outcome.

The async handler additionally swallows the error entirely — `session.rollback(); return;` with no log line at all, unlike its synchronous counterpart.

Async + non-shared is the only combination that behaves correctly: it rolls back without acknowledging, so the messages are redelivered.

## Impact

Silent, unrecoverable message loss whenever writing to a FlowFile fails — a full content repository, a permissions problem, a disk error. The processor logs one line (or none, in async) and continues; the messages are neither in NiFi nor recoverable from Pulsar.

## Suggested direction

Acknowledge only after the content is durably written and the session is committed, and never acknowledge on a path that rolls back. For the Shared case that means deferring the per-message acknowledgement until after the write succeeds rather than issuing it as each message is claimed.

Worth checking `ConsumePulsarRecord` for the same ordering while fixing this — it acknowledges inside `consumeMessages()` in a similar position.

david-streamlio:
Fixes #167.

Both consumers acknowledged messages either **before** their content was written or **after** the session that would have carried them had been rolled back. An acknowledged message is gone from Pulsar — the broker does not redeliver it, and it never reached a FlowFile. Nothing routed to a failure relationship, and in the async path nothing was even logged.

**Change:** both processors now collect the messages they have taken responsibility for and acknowledge them from a `session.commitAsync()` callback — after the content is committed, never before, and never on a path that rolls back. The processors stay at-least-once: a failure after the write means redelivery, which is duplication rather than loss. Async mode keeps using the asynchronous client calls so behaviour matches how it consumed.

**Verification** — new `ConsumePulsarAckOnCommitTest` drives the real processor against a session whose `write()` returns a stream that always throws, which is what a full content repository looks like from here. Against `main`, **both rollback cases fail** (Shared and non-shared), showing messages acknowledged despite the rollback. The healthy-path cases pass on both, so the fix does not simply stop acknowledging.

Full build green: **247 unit + 17 integration**.

---

### Two things found while fixing, worth review attention

**1. Parse failures were being dropped.** My first attempt made the rollback branch skip `handleFailures` entirely, discarding the unparseable messages without routing them. That was my bug, but it exposed that acknowledgement and routing were tangled: parse failures are now routed on **both** branches, and acknowledged only once the `parse_failure` FlowFile has actually been transferred — which `handleFailures` now reports back.

**2. A behaviour decision I would like a second opinion on.** In `ConsumePulsarRecord`, a record set that produced **no records** hit the same branch as a failed write:

```java
} else if (writer != null) {
    // We were able to parse the records, but unable to write them to the FlowFile
    session.rollback();
}
```

Those are different situations. Leaving such messages unacknowledged means the broker redelivers them **forever** — an empty message would loop indefinitely. I made that branch acknowledge, on the grounds that a genuine write failure throws and is routed to `parse_failure` instead, so reaching here means "nothing to emit". The existing `emptyMessageTest` in both sync and async agrees: it expects exactly one acknowledgement. But it is a judgement call about someone else's intent, so flagging it.

### DIFF
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsar.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsar.java
@@ -19,6 +19,7 @@ package org.apache.nifi.processors.pulsar.pubsub;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
@@ -112,6 +113,9 @@ public class ConsumePulsar extends AbstractPulsarConsumerProcessor<byte[]> {
                     Map<String, String> currentAttributes = null;
                     MessageBatchAttributes batchAttributes = null;
 
+                    // acknowledged only once the session carrying their content has committed
+                    final List<Message<GenericRecord>> pendingAcks = new ArrayList<>();
+
                     for (Message<GenericRecord> msg : messages) {
                         currentAttributes = getMappedFlowFileAttributes(context, msg);
 
@@ -130,19 +134,7 @@ public class ConsumePulsar extends AbstractPulsarConsumerProcessor<byte[]> {
                                 session.transfer(flowFile, REL_SUCCESS);
                             }
 
-                            session.commitAsync();
-
-                            if (!shared) {
-	                            final Message<GenericRecord> finalMessage = lastMessage;
-	                            // Cumulatively acknowledge consuming the messages for non-shared subs
-	                            
-	                            getAckService().submit(new Callable<Object>() {
-	                                @Override
-	                                public Object call() throws Exception {
-	                                    return consumer.acknowledgeCumulativeAsync(finalMessage).get();
-	                                }
-	                            });
-                            }
+                            acknowledgeOnCommit(session, consumer, pendingAcks, shared, true);
 
                             lastAttributes = null;
                             lastMessage = null;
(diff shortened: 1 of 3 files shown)


david-streamlio:
Superseded by #169, which is a superset of this and fixes a defect this PR had.

Where they differ, #169 is right and this was wrong:

- **empty record set**: this used `session.rollback()`, which is session-wide and discards FlowFiles already transferred earlier in the same trigger. #169 uses `session.remove()` on just that FlowFile.
- it also catches a variant neither the issue nor this PR identified: on a Shared subscription a message failing schema resolution hit `continue` and was **never acknowledged**, so the broker redelivered it on every ack timeout and it arrived on `parse_failure` repeatedly.
- 24 parameterised cases against this PR's 4, with 17 failing on `main`.

One piece from here is worth keeping and is going up as a follow-up: when the **parse_failure FlowFile itself cannot be written**, #169 rethrows and the caller rolls the whole batch back, so successfully parsed records are redelivered too. This PR left only the unwritable parse-failure messages unacknowledged and kept the acknowledgement for records that were written. Duplicates either way, but far fewer.

Closing in favour of #169.

PR #169, opened 2026-08-28
fix: acknowledge consumed messages only after their FlowFile is committed

### LINKED ISSUE #167: ConsumePulsar acknowledges messages it then discards, losing them on any write error

jselvareef:
## Summary

Closes #167.

`ConsumePulsar` and `ConsumePulsarRecord` told the broker a message was handled either **before** its content was written (Shared/Key_Shared subscriptions acknowledged each message as it was claimed) or **after** rolling the session back on a write error (the cumulative acknowledgement on the other subscription types). An acknowledged message is never redelivered, so any failure to write the FlowFile — a full content repository, a permissions problem, a disk fault — lost the batch for good: neither in NiFi nor recoverable from Pulsar.

**Change**

- New `AbstractPulsarConsumerProcessor.commitAndAcknowledge()`: the acknowledgement runs in the `session.commitAsync()` success callback — the pattern NiFi's own Kafka consumers use — once per committed FlowFile: one acknowledgement per message on Shared/Key_Shared, one cumulative acknowledgement otherwise. A session that is rolled back acknowledges nothing, so the broker redelivers its messages. In asynchronous mode the acknowledgements still go through the acknowledgement service and are drained as before (#150), and the async write error is now logged instead of swallowed.
- Both processors collect the messages of the FlowFile being built in an `uncommitted` list and hand it to `commitAndAcknowledge()` at every point where they used to acknowledge; the acknowledgements on the rollback paths are gone.
- `ConsumePulsarRecord`, which the issue asked to check, had the same ordering plus two more variants:
  - on a Shared subscription a message that failed to resolve a schema was **never** acknowledged (the `continue` skipped the acknowledgement), so the broker redelivered it on every acknowledgement timeout and the flow received it on `parse_failure` again and again. Every message is now acknowledged with the commit of the FlowFile it belongs to, `success` or `parse_failure`;
  - parse failures whose FlowFile could not be written were discarded and acknowledged (#163 left the session clean, but the messages were still lost). `handleFailures()` now removes the FlowFile and rethrows, and the caller rolls the batch back unacknowledged — same outcome as any other write error;
  - an empty record set (`WriteResult.EMPTY`) is discarded with `session.remove()` instead of `session.rollback()`, which also discarded everything routed before it in the session.

Behaviour is otherwise unchanged: same acknowledgement counts, same FlowFiles, same attributes. Under the standard engine `commitAsync()` commits synchronously, so the acknowledgement happens at the same point in the trigger it did before, just after the commit instead of before the write; under the stateless engine it is deferred until the flow completes, which is the correct place for it.

**Verification** — `ConsumePulsarAcknowledgementTest` and `ConsumePulsarRecordAcknowledgementTest`, each parameterised over sync/async × Shared/Exclusive because the four combinations acknowledge through different paths:

- `messagesAreAcknowledgedOnlyAfterTheirFlowFileIsCommitted` — every acknowledgement call records the session state at that moment; the expected value is "1 transferred, committed"
- `nothingIsAcknowledgedWhenTheContentCannotBeWritten` — a `MockProcessSession` whose content streams throw `IOException`, driven through `onTrigger` directly; nothing may be acknowledged and the session has to be clean afterwards
- `discardedEmptyMessagesAreStillAcknowledged` (ConsumePulsar) and `unparseableMessagesAreAcknowledgedOnlyAfterTheirParseFailureFlowFileIsCommitted` (ConsumePulsarRecord) guard the two paths where a message produces no `success` content but still has to be acknowledged

Against `main`, **17 of the 24 cases fail**; the 7 that pass are the combinations the issue already lists as correct (async + non-shared) plus the count-only checks. One caveat: the async Shared ordering case is racy on `main` (the acknowledgement is submitted before the write, and the executor may run it before or after the commit), so it can pass there by chance; with the fix it is deterministic.

Unit suite green: `mvn -B -ntp clean package -Dgpg.skip=true -pl '!docker-image'`, 286 tests. The Testcontainers integration tests were not run here (no Docker on this machine).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Build / CI

## Checklist

- [x] Added or updated tests where it makes sense

### DIFF
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
@@ -630,6 +630,71 @@ public abstract class AbstractPulsarConsumerProcessor<T> extends AbstractProcess
         }
     }
 
+    /**
+     * Commits the session and, once the commit has succeeded, acknowledges {@code messages} to the broker.
+     * <p>
+     * An acknowledged message is gone from the subscription: the broker never redelivers it. So a message
+     * may only be acknowledged once the FlowFile that carries it has been committed. Acknowledging earlier
+     * - as each message was claimed, or on a path that then rolled the session back - told the broker the
+     * message was handled while its content could still be discarded, and a write error (a full content
+     * repository, a permissions problem, a disk fault) lost the message for good: it was neither in NiFi
+     * nor recoverable from Pulsar. The acknowledgement now runs in the commit callback, so a session that
+     * is never committed acknowledges nothing and the broker redelivers its messages instead.
+     * <p>
+     * Shared and Key_Shared subscriptions do not permit cumulative acknowledgements, so every message is
+     * acknowledged individually; the other subscription types acknowledge cumulatively up to the last
+     * message. In asynchronous mode the acknowledgements are submitted to the acknowledgement service and
+     * collected by {@link #drainAcknowledgments()}, as before. {@code messages} is emptied so the caller
+     * can keep collecting the messages of the next commit in the same list.
+     *
+     * @param session  the session holding the FlowFiles the messages were written to
+     * @param consumer the consumer the messages were received from
+     * @param messages the messages carried by the FlowFiles in the session; cleared on return
+     * @param shared   whether the subscription is Shared or Key_Shared
+     * @param async    whether to acknowledge through the asynchronous acknowledgement service
+     */
+    protected void commitAndAcknowledge(final ProcessSession session, final Consumer<GenericRecord> consumer,
+                                        final List<Message<GenericRecord>> messages, final boolean shared, final boolean async) {
+        if (messages.isEmpty()) {
+            return;
+        }
+
+        final List<Message<GenericRecord>> committed = new ArrayList<>(messages);
+        messages.clear();
+
+        session.commitAsync(() -> acknowledge(consumer, committed, shared, async));
+    }
+
+    private void acknowledge(final Consumer<GenericRecord> consumer, final List<Message<GenericRecord>> messages,
+                             final boolean shared, final boolean async) {
+        final ExecutorCompletionService<Object> service = async ? getAckService() : null;
+
+        try {
+            if (shared) {
+                for (final Message<GenericRecord> message : messages) {
+                    if (service != null) {
+                        service.submit(() -> consumer.acknowledgeAsync(message).get());
+                    } else {
+                        consumer.acknowledge(message);
+                    }
+                }
+            } else {
+                final Message<GenericRecord> last = messages.get(messages.size() - 1);
+
+                if (service != null) {
+                    service.submit(() -> consumer.acknowledgeCumulativeAsync(last).get());
+                } else {
+                    consumer.acknowledgeCumulative(last);
+                }
+            }
+        } catch (final PulsarClientException e) {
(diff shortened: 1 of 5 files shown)


jselvareef:
Heads-up: I opened this without noticing #168, which was opened about an hour earlier and takes the same approach (acknowledge from the `session.commitAsync()` callback, nothing on a path that rolls back). Happy to close this one in favour of #168 — your call.

Two things from here that may still be worth carrying over:

- **The tests.** `ConsumePulsarAcknowledgementTest` and `ConsumePulsarRecordAcknowledgementTest` are parameterised over sync/async × Shared/Exclusive, and besides the failing-write case they assert the ordering itself: every acknowledgement records the session state at that moment, and the expected value is `"1 transferred, committed"`. So a regression that acknowledges after the transfer but before the commit is caught too. 17 of the 24 cases fail on `main`.
- **One shared helper.** `commitAndAcknowledge()` lives in `AbstractPulsarConsumerProcessor`, used by both processors, instead of one private copy per processor.

One place where #168 is better than this PR: when the `parse_failure` FlowFile cannot be written, #168 leaves only those messages unacknowledged, whereas this PR rolls the whole batch back, so the successfully parsed records are redelivered as well — duplicates rather than loss, but more of them. I would go with #168's handling there.

If useful, I can rebase the two test classes onto #168 as a follow-up once it lands.

Issue #34, opened 2021-12-21
Add schema-awareness to Pulsar Publisher Processors.


david-streamlio:
Currently, the NiFi processors that publish to Pulsar use the Schema.BYTES schema. They need to make their output match the topics current schema.

tspannhw:
Need to support Schema.JSON, Schema.AVRO.   How to set schemas, send to the topic

PR #172, opened 2026-08-28
feat: encode records with the topic's schema (#34 phase 2)

david-streamlio:
Phase two of #34, stacked on #171. **Review #171 first** — this builds on the producer schema it introduces.

Phase one made the broker validate published content, which stopped unreadable messages reaching a topic. It also left `PublishPulsarRecord` **unable to write to a schema-bearing topic at all**: its Record Writer emits JSON, CSV or Avro-with-header, none of which the broker accepts for an AVRO topic. This closes that.

### Change

New **Message Schema Strategy** property on `PublishPulsarRecord`:

| Strategy | Behaviour |
|---|---|
| `Record Writer` (default) | serialize with the configured writer, exactly as before |
| `Topic Schema` | convert each record to the topic's Avro schema and encode it the way Pulsar does |

`Topic Schema` uses `AvroTypeUtil.createAvroRecord()` for the NiFi→Avro conversion, then writes Avro binary — the encoding Pulsar's AVRO schema uses — so the broker accepts it and a schema-aware consumer decodes it.

**The topic's schema is read from the producer itself.** Producers are created with `Schema.AUTO_PRODUCE_BYTES()`, which binds to the topic on creation and then reports its registered `SchemaInfo`:

```
PROBE34 schemaInfo type=AVRO
PROBE34 schemaInfo def={"type":"record","name":"Sensor","fields":[...]}
```

No admin client, no extra configuration. That was the pivotal unknown and I verified it against a real broker before designing around it.

Default is unchanged, and `Topic Schema` falls back to the Record Writer when the topic has no Avro schema — so enabling it cannot break a flow whose topics are unschematised. That fallback has its own test.

### Tests

`PublishPulsarRecordSchemaIT`: records reaching a schema-bearing topic and decoding through a schema-aware consumer; the default strategy being rejected by such a topic; the fallback on an unschematised topic. Full build green: **268 unit + 30 integration**.

### Two things to know

**1. Only AVRO is encoded.** A JSON-schema topic still needs Record Writer output that happens to match. Worth its own change; I did not want to guess at JSON encoding semantics without a test to pin them.

**2. This surfaced a separate publisher defect.** The round-trip test asserts on the *set* of records, not their order, because they arrive out of order:

```
message 1: id=sensor-1 reading=42
message 2: id=sensor-3 reading=44     <- published second
message 3: id=sensor-2 reading=43     <- published first
```

`PublisherLease` dispatches each send through `CompletableFuture.supplyAsync()`, so records from a single FlowFile race each other. That is pre-existing and unrelated to this change — filed separately.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -49,8 +49,25 @@ jobs:
 
       # Pre-pulling makes the first container start predictable and keeps the pull time out
       # of the per-test startup timeout.
+      # Pulling ahead of time keeps the image download out of the per-test startup timeout. It is an
+      # optimisation, not a requirement - Testcontainers pulls the image itself if this does not - so a
+      # failure here must not fail the build.
+      #
+      # -q -DforceStdout does not guarantee that only the property value reaches stdout: a Maven warning
+      # (a cached dependency-resolution failure, say) lands there too and gets fed to `docker pull`, which
+      # then fails with "invalid reference format". Take the last line and check it looks like an image.
       - name: Pre-pull the Pulsar image
-        run: docker pull "$(mvn -B -ntp help:evaluate -Dexpression=pulsar.image -q -DforceStdout)"
+        continue-on-error: true
+        run: |
+          IMAGE="$(mvn -B -ntp help:evaluate -Dexpression=pulsar.image -q -DforceStdout 2>/dev/null | tail -n 1 | tr -d '[:space:]')"
+          echo "Resolved Pulsar image: '$IMAGE'"
+
+          if [ -z "$IMAGE" ] || printf '%s' "$IMAGE" | grep -qv '^[a-z0-9._/-]\+:[A-Za-z0-9._-]\+$'; then
+            echo "Could not resolve a usable image reference; leaving the pull to Testcontainers."
+            exit 0
+          fi
+
+          docker pull "$IMAGE"
 
       # `verify` is what runs Failsafe; `package` (used by the CI workflow) stops short of it.
       - name: Run integration tests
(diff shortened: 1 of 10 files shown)


david-streamlio:
Applied the cloud review's finding (`merged_bug_001`). It reported **one issue, severity nit, no correctness problems**.

Both halves were legitimate:

- **Per-publish schema re-parsing.** `getTopicAvroSchema()` rebuilt the same Avro type tree on every `publish()`, and leases are pooled, so one lease serving many FlowFiles paid for it repeatedly. Now parsed once.
- **Per-record allocation.** `encodeWithTopicSchema()` allocated a `ByteArrayOutputStream`, `BinaryEncoder` and `GenericDatumWriter` per record, while the surrounding loop already keeps a buffer it resets each iteration for the Record Writer branch. The writer is now built once per record set and the encoder reconfigured via `EncoderFactory.binaryEncoder(out, prior)`, both reusing that buffer.

One deliberate difference from the suggestion: the review proposed caching the parsed schema outright, reasoning that an existing producer's schema cannot change. That is very likely right, but I keyed the cache on the schema definition instead. It costs one string comparison and means a schema that *does* change gets picked up rather than frozen at whatever was seen first — I would rather not bake in an assumption about client internals that I have not verified.

Verified with the integration tests, not just the unit suite: sharing a buffer across records is precisely where two records could concatenate into one message, and the round-trip test publishes several and decodes each through a schema-aware consumer. Full build green: **268 unit + 30 integration**.

Worth noting what the review did *not* flag, since I had called these out as the risk areas: Avro type coverage beyond flat string/int records, and whether plain `GenericDatumWriter` output matches Pulsar's AVRO encoding. The round-trip through a real broker is decent evidence for the second, but neither was independently examined, so they remain the parts most worth a human eye.

PR #179, opened 2026-08-29
fix: start a new record set when the record schema changes

### LINKED ISSUE #174: ConsumePulsarRecord silently drops fields that are not in the first message's schema when the reader infers the schema
`ConsumePulsarRecord` builds the record set writer from the schema of the **first message** of the set and then writes every other message of the batch through it. When the Record Reader infers its schema (the default of `JsonTreeReader`), a field that is not present in that first message is **silently dropped** from every later message of the set, and a field the first message had but a later one lacks is written as `null`. Nothing is logged and nothing is routed to `parse_failure`: the records reach `success` with data missing.

### Reproduction

`ConsumePulsarRecord` on `main` at ccb17ef, `JsonTreeReader` and `JsonRecordSetWriter` from NiFi 2.10.0 with their defaults (*Infer Schema* / *Inherit Record Schema*), synchronous mode, batch size 10, three messages on one topic with no mapped attributes:

```
message 1: {"id":1,"a":"x","nested":{"p":1}}
message 2: {"id":2,"a":"y","extra":"only-in-2","nested":{"p":2,"q":"only-in-2"}}
message 3: {"id":3,"b":true}
```

One FlowFile on `success`, `record.count=3`:

```json
[{"id":1,"a":"x","nested":{"p":1}},
 {"id":2,"a":"y","nested":{"p":2}},
 {"id":3,"a":null,"nested":null}]
```

`extra`, `nested.q` and `b` are gone. Any batch whose first message happens to lack an optional field loses that field for the whole batch, so on a topic with heterogeneous payloads (device telemetry with optional attributes, events of several kinds on one topic) the loss is continuous and invisible.

### Where

`ConsumePulsarRecord.consumeMessages()`:

- the record set boundary is decided on the mapped attributes alone (`!lastAttributes.equals(currentAttributes)`, line 280);
- `schema = getSchema(flowFile, readerFactory, data)` (line 317) resolves the schema from the first message of the set, and `getRecordWriter()` creates the writer with `writerFactory.getSchema(Collections.emptyMap(), srcSchema)` (line 572) — so with *Inherit Record Schema* the writer's schema **is** the first message's inferred schema;
- every message then gets its own reader (line 348), whose schema — the one that actually describes the message — is never looked at. A `RecordSetWriter` writes the fields of the schema it was created with, so extra fields are dropped and missing ones become `null`.

This only bites with inferred schemas: with an explicit schema (Schema Text, a registry) every message resolves to the same schema and unknown fields are dropped by design.

### Impact

Silent data loss on `success`, proportional to how much the payloads vary. It is the kind of loss that is only noticed downstream, when a field is "sometimes missing" in the sink.

### Suggested direction

The reader for each message is created anyway, so its schema is available for free. Two options, in order of ambition:

1. **Treat the schema like the mapped attributes**: when a message's reader schema differs from the schema of the open record set, finish the set and start a new one — `ConsumeKafkaRecord` keys its bundles on `(partition, record schema, attributes)` for exactly this reason. Small change, obviously correct, but heterogeneous payloads may produce many small FlowFiles per batch.
2. **Infer the writer schema from the whole record set rather than its first message** — merge the schemas of the messages that share a set before creating the writer, the way `JsonTreeReader` itself merges the schemas of the records of one input. Keeps one FlowFile per set with every field present (absent ones as `null`, as today), at the cost of a two-pass loop over the batch.

Either way, the README's *Consumer FlowFile attributes* section could carry a note that with an inferred schema the record set's schema currently comes from its first message.

Happy to send a PR for whichever direction you prefer.

jselvareef:
## Summary

Closes #174. Direction (1) from the issue discussion, as agreed there: a change in the record schema closes the record set, the way a change in the mapped attributes already does.

`ConsumePulsarRecord` created a record set's writer from the schema of the set's **first message** and wrote every other message of the set through it. With a reader that infers the schema from the payload — `JsonTreeReader`'s default — any field the first message did not have was silently dropped from the rest of the set, any field it had that a later message lacked came out as `null`, and a field whose type differed between messages made the writer throw `NumberFormatException` out of `onTrigger`. Nothing was logged and nothing reached `parse_failure`.

**Change**

- Each message is parsed **before** the processor decides which record set it joins: the reader is created at the top of the loop (it was created per message anyway, further down) and its schema is compared with the open set's. A different schema finishes the set and starts a new one, exactly like a change in the mapped attributes; an unparseable message has no schema, goes to `parse_failure` and leaves the open set alone.
- The set is opened with the schema of the message that opens it and every message that joins it has the same schema, so the writer never sees a record it cannot represent. `getSchema()` — the separate first-message reader whose schema was the only one consulted — is gone; each reader is closed once its records are written.
- With an explicit schema (Schema Text, a registry, …) every message resolves to the same schema and nothing changes; the explicit-schema test pins the previous output byte for byte.
- The trade-off is documented, as discussed on the issue: the *Consumer Message Batch Size* description now says that `ConsumePulsarRecord` also starts a new FlowFile on a schema change, and the README has a *Record sets and the record schema* section with a "Behaviour change since 2.9.0" callout — a topic with payloads of several shapes produces more, smaller FlowFiles than before, and an explicit schema is the way to keep one FlowFile per batch.

**Verification** — `ConsumePulsarRecordSchemaChangeTest`, running the real `JsonTreeReader` and `JsonRecordSetWriter` because the mock parser has a fixed schema and cannot show any of this:

- `aMessageWithADifferentSchemaStartsANewRecordSet` — the three payloads from #174 plus an array-element field: three FlowFiles, every field kept, nothing invented
- `aTypeConflictStartsANewRecordSetInsteadOfFailingTheTrigger` — `{"v":1}` then `{"v":"text"}`
- `alternatingSchemasProduceOneRecordSetPerChange` — pins the fragmentation trade-off
- `messagesWithTheSameSchemaShareOneRecordSet`, `aMappedAttributeChangeStillStartsANewRecordSet`, `anUnparseableMessageDoesNotCloseTheRecordSet`, `anExplicitSchemaIsAppliedUnchanged` — the behaviour that must not change

Against `main`, **three of the seven fail**:

```
aMessageWithADifferentSchemaStartsANewRecordSet:99->successFlowFiles:212 expected: <3> but was: <1>
aTypeConflictStartsANewRecordSetInsteadOfFailingTheTrigger:123 java.lang.NumberFormatException: For input string: "text"
alternatingSchemasProduceOneRecordSetPerChange:155->successFlowFiles:212 expected: <4> but was: <1>
```

The other four pass on both. No existing test needed a change: the FlowFile counts the suite asserts all come from `MockRecordParser`, whose schema is fixed — checked by running the suite rather than assuming.

Unit suite green: `mvn -B -ntp clean package -Dgpg.skip=true -pl '!docker-image'`, 296 tests. The Testcontainers integration tests were **not** run here (no Docker on this machine); the change does not depend on broker behaviour, so no IT was added.

**Not done here, on purpose:** direction (2) — inferring the set's schema from all of its messages so a heterogeneous batch stays one FlowFile. I have it working on a branch (`DataTypeUtils.merge`, which merges nested records and arrays and widens a type conflict to a choice type rather than picking a winner) and can send it as a follow-up once this lands, if the fragmentation turns out to matter.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Build / CI

## Checklist

- [x] Checked for an existing issue and any open PR already covering this
- [x] Added or updated tests where it makes sense, and confirmed they fail without the change
- [ ] Ran the integration tests (`mvn verify`, needs Docker) — could not: no Docker on this machine

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,21 @@ through *Mapped FlowFile Attributes*.
 > that comes from a message property has to be mapped through *Mapped FlowFile Attributes*
 > instead.
 
+### Record sets and the record schema
+
+`ConsumePulsarRecord` writes consecutive messages with the same mapped attributes (and topic)
+as one record set, using the schema the set was opened with. A message whose schema differs
+starts a new record set - and a new FlowFile - just as a change in the mapped attributes does.
+
+> **Behaviour change since `2.9.0`:** the record set used to be written with the schema of its
+> **first** message. With a Record Reader that infers the schema from each message (the default
+> of `JsonTreeReader`), every field that the first message of a batch did not have was silently
+> dropped from the rest of the batch, and a field whose type differed made the trigger fail. Such
+> batches are now split at every schema change instead, so a topic with payloads of several
+> shapes produces more, smaller FlowFiles than before. If that matters more than the optional
+> fields, give the reader an explicit schema (*Schema Text* or a schema registry): every message
+> then resolves to the same schema and the batch stays one FlowFile.
+
 ## Publisher message metadata
 
 `PublishPulsar` and `PublishPulsarRecord` set the message key and message properties from the
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
@@ -299,7 +299,10 @@ public abstract class AbstractPulsarConsumerProcessor<T> extends AbstractProcess
                     + "default: 1000. If set to a value greater than 1, messages within the FlowFile will be seperated "
                     + "by the Message Demarcator. Consecutive messages are written to the same FlowFile as long as their "
                     + "Mapped FlowFile Attributes are identical; a change in those attributes starts a new FlowFile before "
-                    + "the batch size is reached.")
+                    + "the batch size is reached. ConsumePulsarRecord also starts a new FlowFile when the record schema "
+                    + "changes: with a Record Reader that infers the schema from each message, a batch of messages of "
+                    + "different shapes is split into one FlowFile per shape change rather than losing the fields that "
+                    + "the first message of the batch does not have.")
             .required(false)
             .addValidator(StandardValidators.POSITIVE_INTEGER_VALIDATOR)
             .defaultValue("1000")
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
@@ -19,7 +19,6 @@ package org.apache.nifi.processors.pulsar.pubsub;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.*;
 import java.util.concurrent.BlockingQueue;
@@ -204,7 +203,8 @@ public class ConsumePulsarRecord extends AbstractPulsarConsumerProcessor<Generic
     /**
      * Perform the actual processing of the messages, by parsing the messages and writing them out to a FlowFile.
      * All of the messages passed in shall be routed to either SUCCESS or PARSE_FAILURE, allowing us to acknowledge
-     * the receipt of the messages to Pulsar, so they are not re-sent. The acknowledgement is issued once the
+     * the receipt of the messages to Pulsar, so they are not re-sent. Consecutive messages with the same mapped
+     * attributes and the same record schema share one record set; a change in either starts a new one. The acknowledgement is issued once the
      * session carrying the FlowFiles has been committed; if the batch cannot be written, the session is rolled
      * back and nothing is acknowledged, so the broker redelivers the messages instead of losing them.
      *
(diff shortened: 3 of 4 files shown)


david-streamlio:
Reviewed, and I ran the parts you could not. **Recommend merging.**

### What I verified rather than took on trust

**The integration tests pass.** You noted no Docker on your machine, so this is the gap I could close: merged your branch onto current `main` (which now has #176 and #177) and ran the full build — **277 unit + 30 integration, green**. Nothing in the consumer ITs regressed, including the restart and acknowledgement suites that exercise this class hardest.

**No existing test was modified.** I checked the diff rather than the description: the only test change is the new 237-line file. That was the specific thing I flagged on #174 — several tests assert exact FlowFile counts, and editing one to accommodate a fix is the failure mode the new CONTRIBUTING guide names. You avoided it, and the reason you gave is right: those counts come from `MockRecordParser`, whose schema is fixed.

**Three of seven fail on `main`.** Reproduced by reverting only `ConsumePulsarRecord.java` and running your test:

```
aMessageWithADifferentSchemaStartsANewRecordSet:99 expected: <3> but was: <1>
aTypeConflictStartsANewRecordSetInsteadOfFailingTheTrigger:123 NumberFormatException: For input string: "text"
alternatingSchemasProduceOneRecordSetPerChange:155 expected: <4> but was: <1>
```

### On the change itself

The `NumberFormatException` is a **third** symptom that was not in #174 — the issue described dropped fields and spurious nulls; a type conflict throwing out of `onTrigger` takes the whole trigger down. Worth having found.

Parsing before deciding the set is the right shape, and it is also cheaper than what it replaces: `getSchema()` created a throwaway reader for the first message and then a second reader to read it, where now one reader does both. The reader is closed on every path I could find — parse failure, the `writer == null` branch, and the `finally` after writing.

`alternatingSchemasProduceOneRecordSetPerChange` pinning the fragmentation as a test, rather than only mentioning it in prose, is the right call — it makes the trade-off something a future change has to argue with rather than discover.

### One thing to keep an eye on

Fragmentation now interacts with acknowledgement: each record set commits and acknowledges separately (#169/#170), so a heterogeneous topic produces many small sessions rather than one. Nothing is wrong with that — it is the same at-least-once path either way — but it is a throughput characteristic worth knowing if anyone reports slowness on a mixed-payload topic.

Direction (2) as a follow-up sounds right, and `DataTypeUtils.merge` widening a conflict to a choice type rather than picking a winner answers the objection I raised on the issue about what "merged" means for a type conflict.

PR #182, opened 2026-08-29
fix: apply Message Routing Mode and Max Pending Messages to the producer

### LINKED ISSUE #180: PublishPulsar and PublishPulsarRecord ignore Message Routing Mode and Max Pending Messages
`PublishPulsar` and `PublishPulsarRecord` expose **Message Routing Mode** (`RoundRobinPartition` / `SinglePartition` / `CustomPartition`) and **Max Pending Messages**, but neither property reaches the producer. Whatever the user picks, the producer runs with the Pulsar client defaults.

### Where

`AbstractPulsarProducerProcessor.getPulsarProducerConfiguration()` builds the map that `PublisherPool.createLease()` hands to `ProducerBuilder.loadConf()`. It puts `autoUpdatePartitions`, `autoUpdatePartitionsIntervalSeconds`, `blockIfQueueFull`, `compressionType` and the batching/chunking keys into the map — and nothing else. `MESSAGE_ROUTING_MODE` and `PENDING_MAX_MESSAGES` are read nowhere outside their descriptor definitions.

Both used to be applied. Before 9f478c4 ("Refactored Publish Processors", 2023-05-31) the builder chain had

```java
.maxPendingMessages(context.getProperty(PENDING_MAX_MESSAGES).evaluateAttributeExpressions().asInteger())
.messageRoutingMode(MessageRoutingMode.valueOf(context.getProperty(MESSAGE_ROUTING_MODE).getValue()));
```

The refactor moved producer creation into `PublisherPool` with `loadConf(map)` and those two lines were not carried over into the map. The property descriptors stayed, so the UI still offers the choice. It went unnoticed because the property default, `RoundRobinPartition`, is also the client default.

### Reproduction

Against a real broker (Testcontainers, `apachepulsar/pulsar:4.2.4`): a partitioned topic with 3 partitions, `PublishPulsar` with *Message Routing Mode* = `SinglePartition`, batching disabled, one FlowFile of 30 demarcated, unkeyed messages. Pulsar's `SinglePartition` mode sends unkeyed messages to one randomly chosen partition, so every message should land on the same partition. What arrives:

```
ROUTING MODE SinglePartition -> partitions used: [partition-0, partition-1, partition-2]
```

— round robin, one message per partition in turn. The same flow with the property left at its default produces the identical distribution, which is the tell: the setting changes nothing.

### Impact

- A flow that relies on `SinglePartition` to keep unkeyed messages together (or on `CustomPartition`) silently gets round-robin routing instead, and per-partition ordering it was counting on does not exist.
- *Max Pending Messages* has no effect either, so the producer's pending queue is whatever the client default is rather than the configured bound — relevant with *Async Enabled* and *Block If Queue Full*, which is applied.

Keyed messages are unaffected: a message key is hashed to a partition in every routing mode.

### Fix

Put `messageRoutingMode` and `maxPendingMessages` back into the configuration map — both are `ProducerConfigurationData` fields, so `loadConf` takes them as they are. I intend to send the PR, with a unit test on the configuration map and the integration test above (`SinglePartition` keeps unkeyed messages on one partition; the default still spreads them).

jselvareef:
## Summary

Closes #180.

`PublishPulsar` and `PublishPulsarRecord` expose **Message Routing Mode** and **Max Pending Messages**, but neither reached the producer. They were applied on the `ProducerBuilder` until producer creation moved into `PublisherPool.loadConf(map)` in 9f478c4 (2023-05-31); the map never carried them, so the producer ran with the client defaults whatever the processor was configured with. It went unnoticed because the property default, `RoundRobinPartition`, is also the client default.

**Change**

- `getPulsarProducerConfiguration()` puts `messageRoutingMode` and `maxPendingMessages` into the map — both are `ProducerConfigurationData` fields, so `loadConf()` takes them as they are.
- `CustomPartition` needs a `MessageRouter` handed to the builder, and the processors offer no way to configure one: the client refuses the producer at creation time, which would now turn every trigger into a failure. It is rejected at validation instead, with the reason and the two usable modes. If you would rather drop the option from the allowable values, say so — that is the other defensible fix.
- README: a *Message routing on partitioned topics* section with a "Behaviour change since 2.9.0" callout: a flow that has `SinglePartition` selected really routes unkeyed messages to one partition from now on, *Max Pending Messages* applies, and a flow with `CustomPartition` becomes invalid.

**Verification**

`PublishPulsarRoutingConfigurationTest` asserts the map handed to `loadConf()`, the way the existing chunking test does. Against `main`, all three fail:

```
theConfiguredRoutingModeAndPendingBoundReachTheProducer   expected:<SinglePartition> but was:<null>
theDefaultsAreAppliedExplicitly                           expected:<RoundRobinPartition> but was:<null>
customPartitionIsRejectedAtValidation                     Processor appears to be valid but expected it to be invalid
```

`PublishPulsarRoutingModeIT` publishes 30 demarcated messages to a three-partition topic on a real broker (`apachepulsar/pulsar:4.2.4`) and reads back which partition each one arrived on. Against `main`, `singlePartitionKeepsUnkeyedMessagesOnOnePartition` fails:

```
SinglePartition must route every unkeyed message to the same partition, but they arrived on [partition-0, partition-1, partition-2]
```

`roundRobinSpreadsUnkeyedMessagesOverEveryPartition` and `keyedMessagesStayOnOnePartitionPerKeyInEitherMode` pass on both and guard the behaviour that must not change — in particular that a key is hashed to a partition in either mode, so keyed flows are unaffected by this fix.

Integration tests **were run** this time: `mvn verify -Dgpg.skip=true -Ddocker.api.version=1.41` against a rootless podman socket (`DOCKER_HOST=unix:///run/user/1000/podman/podman.sock`, Ryuk disabled), which the Testcontainers setup from #152 handles without changes. Unit suite: 299 tests, green.

**Not done here:** actual `CustomPartition` support, which would need a way to name a `MessageRouter` implementation; filed nothing for it since the option has never worked and nobody has asked.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Build / CI

## Checklist

- [x] Checked for an existing issue and any open PR already covering this
- [x] Added or updated tests where it makes sense, and confirmed they fail without the change
- [x] Ran the integration tests (`mvn verify`, needs Docker) — via podman, see above

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,20 @@ broker's acknowledgement.
 > **Behaviour change since `2.9.0`:** neither property reached the producer since the publish
 > processors were refactored in 2023 — the producer always ran with the client defaults. A flow
 > that has *Message Routing Mode* set to `SinglePartition` will now really route its unkeyed
-> messages to a single partition, and *Max Pending Messages* now applies. A flow that had
-> `CustomPartition` selected becomes invalid and has to pick one of the other two modes.
+> messages to a single partition. A flow that had `CustomPartition` selected becomes invalid and
+> has to pick one of the other two modes.
+>
+> *Max Pending Messages* now applies **to every flow, including one that never configured it**.
+> The property has always shown a default of `1000`, but the value never reached the client, which
+> ran with the count-based bound disabled. With *Async Enabled*, messages are published through
+> `sendAsync()` and count against that bound; when it is reached and *Block if Message Queue Full*
+> is off — the default — a send fails with `ProducerQueueIsFullError` and its FlowFile is routed to
+> `failure` rather than waiting for room.
+>
+> Whether a flow reaches the bound depends on how fast the broker acknowledges, so it is likelier
+> against a remote or loaded broker than in testing. If you publish large FlowFiles asynchronously,
+> either raise *Max Pending Messages*, set it to `0` to restore the previous unbounded behaviour, or
+> enable *Block if Message Queue Full* so sends wait instead of failing.
 
 ## Publishing to topics that have a schema
 
--- a/README.md
+++ b/README.md
@@ -90,6 +90,22 @@ FlowFile:
 > partition a message routes to, and it makes the topic compactable by that key. If you relied on
 > the previous unkeyed behaviour, clear the `msg.key` attribute before the publish processor.
 
+## Message routing on partitioned topics
+
+*Message Routing Mode* decides where an **unkeyed** message goes on a partitioned topic:
+`RoundRobinPartition` (the default) spreads them over the partitions, `SinglePartition` keeps
+them on one partition chosen per producer. A message with a key is hashed to a partition in
+either mode, so keyed messages keep their order per key regardless of the setting.
+`CustomPartition` needs a `MessageRouter` the processors cannot configure and is rejected at
+validation. *Max Pending Messages* bounds the producer's queue of messages awaiting the
+broker's acknowledgement.
+
+> **Behaviour change since `2.9.0`:** neither property reached the producer since the publish
+> processors were refactored in 2023 — the producer always ran with the client defaults. A flow
+> that has *Message Routing Mode* set to `SinglePartition` will now really route its unkeyed
+> messages to a single partition, and *Max Pending Messages* now applies. A flow that had
+> `CustomPartition` selected becomes invalid and has to pick one of the other two modes.
+
 ## Publishing to topics that have a schema
 
 `PublishPulsar` and `PublishPulsarRecord` create their producers with
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarProducerProcessor.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarProducerProcessor.java
@@ -27,6 +27,8 @@ import org.apache.nifi.annotation.lifecycle.OnStopped;
 import org.apache.nifi.annotation.lifecycle.OnUnscheduled;
 import org.apache.nifi.components.AllowableValue;
 import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.components.ValidationContext;
+import org.apache.nifi.components.ValidationResult;
 import org.apache.nifi.components.Validator;
 import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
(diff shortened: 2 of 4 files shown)


david-streamlio (commented):
Reviewed, and I verified the two claims the change rests on rather than taking them from the description. Both hold up.

**The fix works.** Producer configuration observed on this branch against a real broker:

```
"maxPendingMessages":1000
"messageRoutingMode":"RoundRobinPartition"
```

On `main` the same log line reads `"maxPendingMessages":0` and carries no routing mode at all, so this is the gap closing.

**The `CustomPartition` validation is justified.** I disassembled `ProducerBuilderImpl.setMessageRoutingMode()` in pulsar-client 4.2.4 — with `CustomPartition` selected and a null `customMessageRouter`, it throws `PulsarClientException` at producer creation:

```
72: if_acmpne  104          // routingMode == CustomPartition ?
82: ifnonnull  104          // customMessageRouter == null ?
85: new  PulsarClientException
103: athrow
```

So without the validation every trigger would fail at producer creation with no useful signal. Failing configuration instead is the right call, and the README callout covers the flows it invalidates.

The integration test design is right, too — batching disabled so round robin rotates per message rather than per batching window, and the keyed case run across both modes.

---

### One substantive concern: the new default is a behaviour change for flows that never set the property

*Max Pending Messages* has a descriptor default of `1000`, but the client was running with `0` — the count-based bound disabled. *Block if Message Queue Full* defaults to `false`. With **Async Enabled**, `PublisherLease` publishes through `sendAsync()`, which acquires the pending-message semaphore, so every existing async flow gains a 1000-message ceiling it did not have before, and overflow surfaces as `ProducerQueueIsFullError` rather than blocking.

I probed it: 5000 unbatched async messages in one FlowFile, against the Testcontainers broker.

```
PROBE success=1 failure=0
```

So it does **not** break — a local broker acknowledges fast enough that outstanding sends stay under 1000. I want to be careful not to overstate this: it is not a deterministic regression. But it is a new failure mode that depends on broker latency, and it will show up first on a slow or remote broker under a burst, which is the worst place to discover it.

Two options, either fine by me:

1. Extend the README callout to say that *Max Pending Messages* now takes effect **even for flows that never configured it**, and that async flows publishing large FlowFiles may want *Block if Message Queue Full* enabled.
2. Default the descriptor to `0`, matching what the client has actually been doing, and let users opt into a bound.

I lean slightly to (1) — the property is documented as a bound and users reasonably expect a value in the UI to mean something — but (2) is the more conservative reading of "fix the wiring, do not change behaviour."

### Minor

- `customValidate()` does not call `super.customValidate(validationContext)`. Nothing is lost today, but a superclass validation added later would be dropped silently.
- `theDefaultsAreAppliedExplicitly` asserting `1000` is correct, but it is the *processor's* default, not the client's, which is `0`. Worth a word in the comment so a future reader does not assume they agree.

### Suggestion, not for this PR

`producerConfiguration()` asserting on the map handed to `loadConf()` is exactly the right seam — it is the test that would have caught #180 in 2023. It generalises: a test asserting that **every** declared producer descriptor appears in that map would prevent the next silent drop, since `loadConf` gives no signal for a missing key. I raised that in #186.

Happy for this to go in once the default-change question is settled.

Issue #181, opened 2026-08-29
ConsumePulsarRecord throws NullPointerException on every message from a topic without a schema


jselvareef:
`ConsumePulsarRecord` throws `NullPointerException` out of `onTrigger` for **every** message received from a topic that has no schema. The trigger fails, nothing is acknowledged (#169), and the batch is redelivered to fail again — the processor cannot consume such a topic at all.

### Where

`ConsumePulsarRecord.consumeMessages()`, when it collects the attributes of the record set:

```java
if (msg.getReaderSchema().isPresent() && msg.getReaderSchema().get().getSchemaInfo().getType() == SchemaType.AVRO) {
    currentAttributes.put("avro.schema", new String(msg.getReaderSchema().get().getSchemaInfo().getSchema()));
}
```

The consumer subscribes with `Schema.AUTO_CONSUME()`. For a message from a topic that carries no schema the reader schema **is** present — it is the `AutoConsumeSchema` itself — but its `getSchemaInfo()` returns `null`, and `.getType()` dereferences it:

```
java.lang.NullPointerException: Cannot invoke "org.apache.pulsar.common.schema.SchemaInfo.getType()"
    because the return value of "org.apache.pulsar.client.api.Schema.getSchemaInfo()" is null
```

The `isPresent()` guard was added in #74 (2024-01); it covers "no reader schema" but not "reader schema without schema info", which is what a schema-less topic produces.

### Reproduction

Against a real broker (Testcontainers, `apachepulsar/pulsar:4.2.4`): a topic created with `pulsar-admin topics create-partitioned-topic` (or a plain topic) and never given a schema; 30 JSON messages published; `ConsumePulsarRecord` with `JsonTreeReader`/`JsonRecordSetWriter`, Shared subscription, *Subscription Initial Position* = `Earliest`. Every trigger throws the exception above. With a STRING (or any) schema registered on the topic the same flow consumes all 30 records.

None of the unit tests can show this — the mock message returns an empty reader schema — and no integration test exercises `ConsumePulsarRecord` yet, which is why it survived.

### Fix

Treat a `null` `SchemaInfo` like an absent schema: no `avro.schema` attribute, nothing else changes. I intend to send the PR, with a unit test (mock message whose reader schema has no schema info) and an integration test that consumes from a schema-less topic and from a STRING-schema topic, and reads a partitioned topic as one logical topic on the way.

PR #184, opened 2026-08-29
test: cover the NiFi-to-NiFi round trip

david-streamlio:
## What

Adds `NiFiRoundTripIT` — the first integration test that runs a **publish processor and a consume processor against the same topic**.

Every IT so far exercises one side and arranges or asserts the other with a plain Pulsar client. The pairing a user actually deploys was never covered.

## Why now

This is the gap that let #181 through. `PublishPulsar` creates producers with `Schema.AUTO_PRODUCE_BYTES()`, which validates against an existing schema but **registers nothing** — so a topic this bundle writes to has no schema, and `ConsumePulsarRecord` threw `NullPointerException` on every message from one.

Verified against a real broker:

```
### produce with the DEFAULT (bytes) schema
"msgInCounter" : 3,
--- schemas get persistent://public/default/bytes-producer ---
Schema not found
```

The fix in #183 is correct, but its own IT arranges the topic with a raw `Schema.BYTES` producer — so the pairing that actually broke stayed uncovered. This closes that.

## Tests

| Round trip | Catches #181 |
|---|---|
| `PublishPulsar` → `ConsumePulsar` | no |
| `PublishPulsar` → `ConsumePulsarRecord` | yes |
| `PublishPulsarRecord` → `ConsumePulsarRecord` | yes |

The record tests assert up front that the topic carries no schema, so if `AUTO_PRODUCE_BYTES` ever starts registering one they fail loudly rather than quietly ceasing to cover the case they were written for.

## Failing first

With `ConsumePulsarRecord` reverted to its pre-#183 state:

```
NiFiRoundTripIT.recordsAreReadBackFromATopicThisBundleWrote  NullPointerException
NiFiRoundTripIT.recordsSurviveTheRecordToRecordRoundTrip     NullPointerException
Tests run: 3, Failures: 2
```

With the fix in place:

```
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
```

The bytes round trip passes either way — `ConsumePulsar` never reads the reader schema — which is why the record pairings are the ones carrying the coverage.

## Not covered

- Schema-bearing topics end to end. `PublishPulsarRecord` encodes AVRO and JSON against a registered schema, but nothing reads that back through `ConsumePulsarRecord`.
- Message keys and properties surviving the round trip.

### DIFF
--- a/nifi-pulsar-processors/src/test/java/org/apache/nifi/processors/pulsar/it/NiFiRoundTripIT.java
+++ b/nifi-pulsar-processors/src/test/java/org/apache/nifi/processors/pulsar/it/NiFiRoundTripIT.java
@@ -1,0 +1,221 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.pulsar.it;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.TreeSet;
+
+import org.apache.nifi.json.JsonRecordSetWriter;
+import org.apache.nifi.json.JsonTreeReader;
+import org.apache.nifi.processors.pulsar.AbstractPulsarConsumerProcessor;
+import org.apache.nifi.processors.pulsar.AbstractPulsarProducerProcessor;
+import org.apache.nifi.processors.pulsar.pubsub.ConsumePulsar;
+import org.apache.nifi.processors.pulsar.pubsub.ConsumePulsarRecord;
+import org.apache.nifi.processors.pulsar.pubsub.PublishPulsar;
+import org.apache.nifi.processors.pulsar.pubsub.PublishPulsarRecord;
+import org.apache.nifi.reporting.InitializationException;
+import org.apache.nifi.util.MockFlowFile;
+import org.apache.nifi.util.TestRunner;
+import org.apache.nifi.util.TestRunners;
+import org.junit.Test;
+import org.testcontainers.containers.Container;
+
+/**
+ * NiFi in, NiFi out: what this bundle publishes, this bundle can read back.
+ * <p>
+ * Every other integration test here exercises one side and arranges or asserts the other with a plain
+ * Pulsar client. That leaves the pairing itself untested, and the pairing is what a user actually
+ * deploys. It is also where #181 lived: {@code PublishPulsar} creates producers with
+ * {@code Schema.AUTO_PRODUCE_BYTES()}, which does <em>not</em> register a schema on a topic that had
+ * none, so a topic written by this bundle is a schema-less topic - and {@code ConsumePulsarRecord}
+ * threw {@code NullPointerException} on every message from one. A raw-client test on either side alone
+ * could not see it; publishing with {@code Schema.BYTES} and consuming with {@code Schema.STRING} both
+ * look fine in isolation.
+ */
+public class NiFiRoundTripIT extends AbstractPulsarIT {
+
+    private static final int MESSAGES = 20;
+
+    /** {@code PublishPulsar} to {@code ConsumePulsar}: the plainest pairing, raw bytes both ways. */
+    @Test
+    public void bytesSurviveTheRoundTrip() throws Exception {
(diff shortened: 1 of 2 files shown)


david-streamlio:
@jselvareef would you be willing to take a look at this one?

It grew out of your #183. While verifying that fix I checked what `PublishPulsar` actually leaves on a topic, and `AUTO_PRODUCE_BYTES` registers no schema — so a topic this bundle writes to *is* the schema-less topic from #181:

```
### produce with the DEFAULT (bytes) schema
"msgInCounter" : 3,
--- schemas get persistent://public/default/bytes-producer ---
Schema not found
```

Your IT arranges the topic with a raw `Schema.BYTES` producer, which is a faithful reproduction, but it means the publish/consume pairing that actually breaks was still untested. This adds that: `NiFiRoundTripIT` (bytes, record, and record-to-record) and `NiFiSchemaRoundTripIT` (AVRO and JSON round trips, plus both mismatched reader/topic combinations).

With `ConsumePulsarRecord` reverted to its pre-#183 state, 2 of the 3 round trips fail with the NPE and the bytes one passes — which is the part I'd most like a second opinion on, since it means `ConsumePulsar` coverage could never have caught this.

Two things you'll know better than I do:

1. The AVRO round trip only works with an `AvroReader` whose Schema Text is `${avro.schema}`, because Pulsar writes bare Avro binary with no embedded schema. Is that the configuration you'd expect users to land on, or is the `avro.schema` attribute meant to be consumed some other way?
2. I've written the mismatch behaviour — everything to `parse.failure`, nothing to `success`, batched into one demarcated FlowFile — as *correct*. If that was a deliberate design choice rather than a consequence, it'd be worth saying so in the test.

Also filed #185 off the back of it: using the `GenericRecord` that `AUTO_CONSUME` already decodes, so the reader wouldn't need to match the topic's encoding at all. Interested in whether that seems worth doing to you.

jselvareef:
Happy to. Looked at both ITs against the code and the history; here is what I can add.

**1. The bytes round trip could not have caught #181, and never will.** `ConsumePulsar` never calls `getReaderSchema()` — only `ConsumePulsarRecord` does, for the `avro.schema` attribute — so the record pairings are the right carriers for anything derived from the reader schema, now or later.

**2. `AvroReader` with Schema Text `${avro.schema}` is the intended configuration, not a workaround.** #71 says so: with no external registry the only option is the *Schema Text Property* strategy, and NiFi's `AvroReader` ships `${avro.schema}` as the default of that field — which is why the attribute got that name. Two things make it hold up: the attribute is part of the record set's attribute key, so a schema version change opens a new FlowFile carrying its own schema; and since #179 the reader is created with those same attributes, so `${avro.schema}` resolves for the reader itself, not only downstream. What is missing is a README line saying exactly that — the attribute table only says "the topic schema, when it is an AVRO schema". Happy to add it, or it fits your #191 README changes.

**3. The mismatch behaviour is the contract.** `parse_failure`, `handleFailures()` and the demarcated batching are in the initial commit, and the Javadoc states the rule: every message is routed to success or parse_failure so that it can be acknowledged. #169/#170 rely on it — the parse_failure FlowFile is what carries those messages to their acknowledgement. Nothing partially parsed reaches success, and the raw payload is recoverable from parse_failure. Worth saying in the test as you suggest.

**4. #185 is worth doing, as an opt-in strategy with the reader as the default** — which is what #191 does. It removes the undiscoverable AvroReader configuration and the whole encoding-mismatch class of failures, and schema evolution comes for free. The reader has to stay the default: topics with no schema (the ones this bundle itself writes, per your finding), JSON inference, CSV, and primitive topics — STRING topics with JSON text inside are common in production and #189 should keep "let the reader parse the STRING payload" possible.

david-streamlio:
Thank you — and apologies, I merged this about seven minutes before your comment landed, so it went in without the benefit of it. Everything you raised has been acted on rather than lost; here is where each point went.

**2. The AvroReader configuration.** You are right, and I had it weaker than the truth. I checked `SchemaAccessUtils` and `SCHEMA_TEXT` really does default to `${avro.schema}` with `FLOWFILE_ATTRIBUTES` scope:

```
403: ldc_w  #282   // String ${avro.schema}
406: invokevirtual // PropertyDescriptor$Builder.defaultValue
```

So only the *Schema Access Strategy* needs changing, not the schema text — my README line implied both. Fixed in #191 (`2b1d54d`), along with the two properties you named that were nowhere in the docs: that `avro.schema` participates in the record-set boundary so a version change opens a new FlowFile with its own schema, and that since #179 the reader is created with the message's attributes so `${avro.schema}` resolves for the reader itself. Both are the reason it works at all, and neither was written down.

**3. The parse.failure contract.** Taken, and this is the change I am most glad you pushed for. The test asserted the behaviour without saying it was deliberate, so a later reader could reasonably have taken it as incidental and "fixed" it. The class comment now states the rule — every message routed to `success` or `parse.failure` so it can be acknowledged, the `parse.failure` FlowFile being what carries undecodable messages to their acknowledgement, and #169/#170 depending on that.

**4. Primitive topics.** This is the one I would have got wrong. A `STRING` topic carrying JSON text wanting the reader to parse it had not occurred to me, and #189 as filed would have made "primitive schema" mean "one-field record" unconditionally. I have added a section to #189 saying the reader must stay reachable for those topics, and flagging that the likely resolution — fall back to the reader when one is configured, single-field record when not — means the same topic behaves differently depending on whether a reader is set, which should be a deliberate decision rather than something we back into.

**1.** Agreed, and it is the reason the bytes round trip is in the suite at all: it is the control that shows why the record pairings are the ones carrying the coverage.

On the README line for the attribute table — I have taken the wording into #191 since it was already editing that section, so please do not duplicate the effort. #191 is the #185 implementation and follows your reading exactly: opt-in strategy, reader stays the default, and the reader remains the fallback for schema-less topics. Review very welcome there if you have the time.

Issue #190, opened 2026-08-30
Support KeyValue schema topics in the record processors


david-streamlio:
## Ask

Support Pulsar's [KeyValue schema](https://pulsar.apache.org/docs/5.0.x/schema-understand/#keyvalue-schema) in `PublishPulsarRecord` and `ConsumePulsarRecord`.

A `KEY_VALUE` topic carries **two** schemas — one for the key, one for the value — and an encoding that says where the key is written:

- **`INLINE`** — key and value are both in the message payload.
- **`SEPARATED`** — the key goes in the message's key metadata and only the value is in the payload.

## Where it stands today

`SchemaType.KEY_VALUE` is unrecognised on both sides. `PublisherLease.getTopicSchema()` returns null for it, so records are serialized with the configured Record Writer, and the broker then validates that output against the topic's KeyValue schema. The `Topic Schema` consume strategy from #185 likewise falls back to the Record Reader, which sees the raw payload with no idea a key schema exists.

The bundle does have a `msg.key` concept already — `MESSAGE_KEY` on publish, `PULSAR_MESSAGE_KEY` in *Mapped FlowFile Attributes* on consume — but it is a plain string, unrelated to schemas. A `SEPARATED` KeyValue topic puts a **schema-encoded** key in that same metadata field, so the two features meet and the interaction needs deciding.

## The design questions

1. **Record shape.** The natural NiFi mapping is a record with two nested record fields, `key` and `value`, each shaped by its own schema. Worth confirming against what a NiFi user would expect to write in a downstream query.
2. **Which encoding to support first.** `SEPARATED` is the more common in practice — it is what makes a topic compactable by key — but it is also the one that collides with the existing `msg.key` handling.
3. **The collision.** On a `SEPARATED` topic, does *Message Key* still apply, or is the key taken from the record's `key` field? Both cannot be true. Most likely: the KeyValue schema wins for such topics and *Message Key* is rejected at validation, in the same shape as the `CustomPartition` rejection in #182.
4. **Consuming.** Should the key be surfaced only inside the record, or also as FlowFile attributes so it can be routed on without parsing? The existing mapped-attribute mechanism gives the key as a string, which is not meaningful for an encoded key.

## Suggested scope

Decide the record shape and the `msg.key` interaction first — they are the parts that are expensive to change later. Then `SEPARATED` with an AVRO or JSON key schema, then `INLINE`.

## Related

[...]

Issue #189, opened 2026-08-30
Support primitive schema topics in the record processors


david-streamlio:
## Ask

Handle topics whose registered schema is a **primitive** type — `STRING`, `INT8`/`INT16`/`INT32`/`INT64`, `FLOAT`, `DOUBLE`, `BOOLEAN`, `BYTES`, and the date/time family (`DATE`, `TIME`, `TIMESTAMP`, `INSTANT`, `LOCAL_DATE`, `LOCAL_TIME`, `LOCAL_DATE_TIME`) — in `PublishPulsarRecord` and `ConsumePulsarRecord`.

## Where it stands today

Both sides recognise exactly two schema types and treat everything else as "no schema".

Publishing, in `PublisherLease.getTopicSchema()`:

```java
if (info == null || (info.getType() != SchemaType.AVRO && info.getType() != SchemaType.JSON)) {
    return null;   // falls back to the configured Record Writer
}
```

Consuming, in the `Topic Schema` strategy added for #185, the same two types are supported and anything else falls back to the Record Reader.

So for a topic with, say, a `STRING` schema, the processors emit whatever the Record Writer produces. The broker validates it against the topic's schema (`AUTO_PRODUCE_BYTES`, #34), and it is accepted only if the writer's output happens to match — a CSV writer's `id, reading` line is valid UTF-8 and will be *accepted* by a `STRING` topic while being meaningless as a value.

## The design question

A primitive topic carries **one value per message**, not a record with fields, so the mapping to NiFi's record model has to be decided rather than derived:

- Consuming: wrap the value in a single-field record. The field name needs choosing — `value` is the obvious candidate — and it should probably be configurable, since it becomes a column name downstream.
- Publishing: the inverse, and it needs a rule for what happens when the incoming record has more than one field. Reject, or take a named field?

The date/time types are worth calling out separately: they map onto Avro logical types and NiFi's `RecordFieldType.TIMESTAMP` / `DATE` / `TIME`, so they are more than a passthrough and deserve their own tests. Note that logical types are **not** covered even for AVRO today (see #34).

## The reader must stay reachable for primitive topics

From review feedback on #184: a `STRING` topic whose messages contain JSON text is a common production
shape. Those flows want the payload **parsed by a Record Reader** into a real record, not wrapped as a
single string field — so "the topic has a primitive schema" must not automatically mean "emit a one-field
record".

Whatever this issue adds therefore has to leave the reader path reachable for such topics. The likely
[...]

PR #191, opened 2026-08-30
feat: build records from the topic's schema instead of a Record Reader (#185)

### LINKED ISSUE #185: Support Pulsar generic struct schemas (GenericRecord) end to end
## Ask

Support Pulsar's [generic schema API](https://pulsar.apache.org/docs/schema-understand/) — `RecordSchemaBuilder` / `Schema.generic(schemaInfo)` / `GenericRecord` — as a first-class way to move structured data through these processors, rather than requiring the user to hand-match a NiFi record reader to whatever the topic happens to be encoded as.

## Why

`ConsumePulsarRecord` subscribes with `Schema.AUTO_CONSUME()`, which decodes each message into a `GenericRecord`. But the processor then ignores the decoded value and passes the **raw wire bytes** to the configured Record Reader:

```java
final byte[] data = msg.getData();
reader = readerFactory.createRecordReader(currentAttributes, new ByteArrayInputStream(data), data.length, getLogger());
```

So the reader has to match the topic's encoding exactly, and the two encodings are not interchangeable:

| Topic schema | On the wire | Reader that works |
|---|---|---|
| AVRO | bare Avro binary, no embedded schema | `AvroReader` with schema access `schema-text-property` and Schema Text `${avro.schema}` |
| JSON | JSON text | `JsonTreeReader` |
| none | whatever was published | any reader that fits the payload |

That AVRO configuration is not discoverable. The schema is never in the data — Pulsar keeps it in the registry — so the default "embedded avro schema" strategy cannot work, and the `avro.schema` attribute that makes it possible is only mentioned in a code comment.

Verified against a real broker (`apachepulsar/pulsar:4.2.4`): pointing the wrong reader at a topic sends **every message** to `parse.failure`, batched into a single FlowFile, with nothing reaching `success`:

```
PROBE success=0 parse_failure=1
PROBE   pf len=121
```

No data is lost and nothing gets stuck — the handling is correct — but the flow silently produces zero records until someone works out that the reader and the topic disagree.

## Proposal

Let the topic's own schema drive record construction, mirroring `PublishPulsarRecord`'s existing **Message Schema Strategy** property:

- Add a `Topic Schema` strategy to `ConsumePulsarRecord` that builds NiFi `Record`s from `msg.getValue()` (the `GenericRecord` `AUTO_CONSUME` already produced), mapping the Pulsar `SchemaInfo` to a NiFi `RecordSchema`. With that selected, no Record Reader needs configuring at all, and AVRO and JSON topics behave identically.
- Keep the current reader-driven behaviour as the default, so existing flows are unaffected.

## Scope note

The publish side has the same boundary and it is worth deciding together. `PublisherLease.getTopicSchema()` returns a schema only for AVRO and JSON:

```java
if (info == null || (info.getType() != SchemaType.AVRO && info.getType() != SchemaType.JSON)) {
```

Every other type — `KEY_VALUE`, `PROTOBUF_NATIVE`, and the primitives — falls back to the configured Record Writer, whose output the broker then rejects unless it happens to match. Generic support on the consume side would leave that asymmetry in place.

## Related

- #34 — publisher schema awareness, phases 1–3 (AVRO and JSON encoding) merged
- #181 / #183 — schema-less topics, where the reader-driven design also showed through
- #184 — the NiFi-to-NiFi round-trip tests; the per-schema-type suite there is what surfaced this

david-streamlio:
Closes #185.

## The problem

`ConsumePulsarRecord` subscribes with `Schema.AUTO_CONSUME()`, which decodes every message — and then throws the decoded value away, passing the **raw wire bytes** to the configured Record Reader. So the reader has to match the topic's encoding, and the two encodings are not interchangeable:

| Topic | On the wire | Reader that works |
|---|---|---|
| AVRO | bare Avro binary, no embedded schema | `AvroReader`, Schema Text `${avro.schema}` |
| JSON | JSON text | `JsonTreeReader` |

The AVRO configuration is not discoverable. The schema is never in the data — Pulsar keeps it in the registry — so `AvroReader`'s default "embedded avro schema" strategy cannot work, and the `avro.schema` attribute that makes it possible was only mentioned in a code comment. Pointing the wrong reader at a topic sends **every** message to `parse.failure`.

## The change

A **Message Schema Strategy** property, mirroring the one `PublishPulsarRecord` already has:

- **`Record Reader`** (default) — exactly today's behaviour. No existing flow changes.
- **`Topic Schema`** — records are built from the schema the topic carries. No Record Reader, no schema configuration of any kind, and AVRO and JSON behave identically.

The field definitions come from the **broker**, not the payload and not message properties: Pulsar keeps schemas in a registry keyed by topic and version, and `AUTO_CONSUME` attaches each message's schema as its reader schema. Because it arrives per message, **schema evolution needs no special handling** — a message published under an older version decodes with the version it was written with, and #179's record-set boundaries already close a set when the schema changes.

### Why it decodes the bytes itself

The obvious implementation — use the `GenericRecord` Pulsar already decoded — does not work. The client **shades Avro**, so `msg.getValue()` is backed by `org.apache.pulsar.shade.org.apache.avro...`, which NiFi's `AvroTypeUtil` cannot accept. Parsing the definition with unshaded Avro and decoding here is the exact mirror of `PublisherLease`'s encoding, and keeps both directions on the same conversion code.

### The Record Reader descriptor

It became optional, with the requirement moved into `customValidate`. Leaving it required would make the new strategy unusable; making it optional outright would silently accept a reader-driven consumer with no reader, which has always been invalid. It is still the fallback under `Topic Schema` for topics the strategy cannot decode.

## Failing first

With `TopicSchemaRecordDecoder.supports()` forced to `false`:

```
Tests run: 6, Failures: 4
```

As implemented:

```
Tests run: 6, Failures: 0
```

The two that pass either way are the fallback cases, which is the intended discrimination — they must not depend on the decoder.

Full suite: **281 unit tests and 44 integration tests green**, confirming the default reader path is unchanged.

## Tests

- `TopicSchemaRecordDecoderTest` — AVRO binary and JSON text decoding, JSON type coercion, re-parsing on a changed definition, and which schema types are supported.
- `ConsumePulsarRecordSchemaStrategyTest` — the reader stays required under the default strategy, optional under `Topic Schema`.
- `ConsumePulsarTopicSchemaIT` — against a real broker: AVRO and JSON consumed with no reader configured, plus the schema-less fallback and the no-reader `parse.failure` case.
- **e2e:** publish with `PublishPulsarRecord` on `Topic Schema` and read it back with `ConsumePulsarRecord` on `Topic Schema`, with no Record Reader anywhere in the flow, on both encodings.

## Not covered

- **Primitive schemas (#189)** and **KeyValue schemas (#190)** — both fall back to the Record Reader; filed separately.
- **Logical types** — timestamps, decimals, UUIDs are untested in either direction (also open under #34).
- **Schema evolution against a live broker** — exercised only through a changed definition in the decoder's unit test, not by evolving a topic's registered schema mid-stream.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,29 @@ broker's acknowledgement.
 > either raise *Max Pending Messages*, set it to `0` to restore the previous unbounded behaviour, or
 > enable *Block if Message Queue Full* so sends wait instead of failing.
 
+## Consuming from topics that have a schema
+
+`ConsumePulsarRecord`'s **Message Schema Strategy** decides how a message becomes records.
+
+`Record Reader` (the default) parses each message with the configured reader, which has to match how the
+topic is encoded. That is not always obvious: an AVRO topic carries *bare Avro binary* — no file header
+and no embedded schema, because Pulsar keeps the schema in its registry — so it needs an `AvroReader`
+whose *Schema Access Strategy* is `Use 'Schema Text' Property` and whose *Schema Text* is
+`${avro.schema}`, the attribute this processor sets from the topic's registered schema. A JSON topic
+carries text a `JsonTreeReader` can infer. Pointing the wrong reader at a topic sends every message to
+`parse.failure`.
+
+`Topic Schema` builds records from the schema the topic carries instead. The field definitions come from
+the broker — Pulsar attaches each message's schema to it — so no reader and no schema configuration are
+needed at all, and AVRO and JSON topics behave identically. Because the schema arrives per message,
+evolution is handled: a message published under an older version decodes with the version it was written
+with.
+
+A *Record Reader* may still be configured under `Topic Schema`, and messages the strategy cannot decode
+fall back to it — a topic with no schema at all, or one whose schema has no record shape. Without a
+reader to fall back to, those messages go to `parse.failure`. Primitive and `KeyValue` schemas are not
+yet decoded this way.
+
 ## Publishing to topics that have a schema
 
 `PublishPulsar` and `PublishPulsarRecord` create their producers with
--- a/README.md
+++ b/README.md
@@ -125,10 +125,13 @@ broker's acknowledgement.
 `Record Reader` (the default) parses each message with the configured reader, which has to match how the
 topic is encoded. That is not always obvious: an AVRO topic carries *bare Avro binary* — no file header
 and no embedded schema, because Pulsar keeps the schema in its registry — so it needs an `AvroReader`
-whose *Schema Access Strategy* is `Use 'Schema Text' Property` and whose *Schema Text* is
-`${avro.schema}`, the attribute this processor sets from the topic's registered schema. A JSON topic
-carries text a `JsonTreeReader` can infer. Pointing the wrong reader at a topic sends every message to
-`parse.failure`.
+with *Schema Access Strategy* set to `Use 'Schema Text' Property`. Its *Schema Text* already defaults to
+`${avro.schema}`, which is the attribute this processor sets from the topic's registered schema, so the
+access strategy is the only field that has to change. Because `avro.schema` is part of what decides a
+record set's boundaries, a schema version change closes the current FlowFile and opens a new one carrying
+its own schema, and the reader is created with those attributes so `${avro.schema}` resolves for the
+reader itself rather than only downstream. A JSON topic carries text a `JsonTreeReader` can infer.
+Pointing the wrong reader at a topic sends every message to `parse.failure`.
 
 `Topic Schema` builds records from the schema the topic carries instead. The field definitions come from
 the broker — Pulsar attaches each message's schema to it — so no reader and no schema configuration are
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
@@ -37,7 +37,10 @@ import org.apache.nifi.annotation.behavior.WritesAttributes;
 import org.apache.nifi.annotation.documentation.CapabilityDescription;
 import org.apache.nifi.annotation.documentation.SeeAlso;
 import org.apache.nifi.annotation.documentation.Tags;
+import org.apache.nifi.components.AllowableValue;
 import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.components.ValidationContext;
+import org.apache.nifi.components.ValidationResult;
 import org.apache.nifi.expression.ExpressionLanguageScope;
 import org.apache.nifi.flowfile.FlowFile;
 import org.apache.nifi.processor.ProcessContext;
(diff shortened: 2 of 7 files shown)


jselvareef (commented):
Reviewed against the code and ran the integration tests locally — `ConsumePulsarTopicSchemaIT`, `ConsumePulsarRecordIT`, `NiFiRoundTripIT`, `NiFiSchemaRoundTripIT`: 16/16 green (podman), 292 unit tests green. The design is the one from the issue — opt-in, reader stays the default and the fallback — and decoding the bytes here rather than using the shaded `GenericRecord` is the right call. Two findings, both reproduced, one of them serious.

**1. `customValidate()` drops the parent's validation.** `ConsumePulsarRecord.customValidate()` starts from a fresh list and never calls `super.customValidate(validationContext)`, so none of `AbstractPulsarConsumerProcessor`'s rules apply to this processor any more. Probe with a `TestRunner` (reader, writer and subscription set):

```
main: no Topics and no Topics Pattern → valid=false      Ack Timeout = 2 sec → valid=false
#191: no Topics and no Topics Pattern → valid=true       Ack Timeout = 2 sec → valid=true
```

A `ConsumePulsarRecord` with no topic at all is now a valid configuration. Fix: seed `results` with `super.customValidate(validationContext)` and add the reader rule to it. (Same shape you flagged on #182 — there nothing sat behind `super`; here everything the consumer validates does.)

**2. The shared `TopicSchemaRecordDecoder` is not thread-safe, and the failure is silent.** The decoder is a field of the processor, so every concurrent task shares its four cache fields (`cachedDefinition`, `cachedType`, `cachedAvroSchema`, `cachedRecordSchema`), which are read and replaced with no synchronization. Two triggers decoding different schemas at the same time — two topics in *Topics*, a schema version change mid-stream, or just *Concurrent Tasks* > 1 while a topic evolves — tear the cache, and one thread decodes with the other's Avro schema. Probe: one decoder, two threads, two single-field AVRO schemas (`A{alpha}` / `B{beta}`), 20,000 decodes each:

```
wrongRecords=14360   exceptions=0   of 40000 decodes
```

36% of the records came back with the wrong shape or value and **no exception** — the writer would have sent them to `success`. `TopicSchemaRecordDecoderTest` cannot see it because it is single-threaded. Options: (a) create the decoder per `consumeMessages()` call — the definition is then parsed once per batch, cheap, and nothing needs locking; (b) keep the processor-level cache but hold an immutable snapshot (`ParsedSchema{definition, type, avroSchema, recordSchema}`) behind one `volatile` reference, swapped atomically, so a reader always sees a consistent set; (c) a small `ConcurrentHashMap<definition, ParsedSchema>` so several topics or versions do not thrash. Whichever you pick, a two-thread / two-schema test like the probe would pin it.

**Minor**

- `decode()` builds `new String(schemaInfo.getSchema())` and compares it on every message; comparing the bytes first (`Arrays.equals`, or identity — `AutoConsumeSchema` hands back the same `SchemaInfo` per version) skips a per-message allocation of the whole definition.
- `decodeJson` uses `dropUnknownFields = true`, so payload fields absent from the registered JSON schema vanish silently. Defensible — the schema is the contract — but worth a line in the README, since under *Record Reader* with inference they survive.
- `getLastRecordSchema()` has no caller.
- README and test text say `parse.failure`; the relationship is `parse_failure`.

Everything else reads right: the record-set boundary keyed on `currentSchema` whichever path produced it, decode failures landing on `parse_failure`, and the STRING-topic fallback matching what #189 needs.

PR #192, opened 2026-08-30
feat: map primitive-schema topics to single-field records (#189)

### LINKED ISSUE #189: Support primitive schema topics in the record processors

david-streamlio:
Closes #189.

## The change

A topic whose registered schema is a primitive — `STRING`, `BOOLEAN`, `INT8`/`INT16`/`INT32`/`INT64`, `FLOAT`, `DOUBLE` — carries one value per message and has no fields. Both processors previously treated that as no schema at all: the Record Writer serialized whatever it was given and the broker accepted it only if it happened to match.

Now:

- **Consuming** — `Topic Schema` gives each message a record of one field, named by a new **Primitive Value Field** property (`value` by default). It becomes a column name downstream, so it is configurable.
- **Publishing** — a single-field record is encoded with the topic's primitive schema, the value coerced to its type.

Encoding and decoding go through Pulsar's own `Schema.STRING`, `Schema.INT32` and friends rather than reimplementing the wire formats the broker validates against.

## The precedence rule, from review feedback on #184

> STRING topics with JSON text inside are common in production and #189 should keep "let the reader parse the STRING payload" possible.

So **a configured Record Reader wins on a primitive topic.** With a reader, the payload is parsed into real records; without one there is nothing to parse with, so the single-field record is the useful answer rather than a parse failure. Both directions are tested.

This does mean the same topic behaves differently depending on whether a reader is configured. That is deliberate and documented, rather than something to discover.

## Two things the tests caught

**`BYTES` cannot be supported, and an IT proved it.** `AUTO_PRODUCE_BYTES` reports a topic with *no* schema as `BYTES` with an empty definition:

```
info={"name": "Bytes", "schema": "", "type": "BYTES"} type=BYTES
```

A registered `BYTES` schema is therefore indistinguishable from none. Treating it as primitive made every schema-less topic reject multi-field records — and schema-less is exactly what this bundle's own publishers produce (#181). It is excluded, with the reasoning recorded in the code.

**An unchecked exception would have escaped the publish loop.** `DataTypeUtils.toInteger` throws `NumberFormatException`, not `IllegalTypeConversionException`, for a String that is not a number. My original catch missed it, so it would have failed the whole trigger instead of routing that FlowFile to `failure`.

## Publishing guard

A primitive topic takes one value per message, so a record with several fields has no unambiguous mapping. Guessing which field was meant would publish the wrong data silently, so it routes to `failure` with a message naming the fields.

## Tests

- `PrimitiveTopicSchemaTest` — every encode is checked against **Pulsar's own decoder**, not against our encoder, so a test cannot pass by being wrong in both directions. Plus coercion, malformed payloads, and which types are supported.
- `PrimitiveSchemaRoundTripIT` — against a real broker: **e2e NiFi→NiFi round trips on STRING and INT32**, the configurable field name, both sides of the reader-precedence rule, the multi-field publish guard, and a schema-less topic being unaffected.

**299 unit tests and 58 integration tests green.**

## Not covered

- The **date and time schemas** (`DATE`, `TIME`, `TIMESTAMP`, `INSTANT`, `LOCAL_DATE`, `LOCAL_TIME`, `LOCAL_DATE_TIME`), deferred for the same reason AVRO logical types are still uncovered under #34 — they carry conversion questions the numerics do not and deserve their own tests.
- **`KeyValue` schemas**, which are #190.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,10 +151,15 @@ value per message and has no fields, so `Topic Schema` gives each message a reco
 named by **Primitive Value Field** (`value` by default). It becomes a column name downstream, so it is
 worth setting to something meaningful.
 
-**A configured *Record Reader* takes precedence on these topics.** A `STRING` topic carrying JSON text is
-a common shape, and those flows want the payload parsed into records rather than wrapped in one string
-field — so with a reader configured the payload is parsed, and the single-field record is what happens
-when there is no reader to parse with.
+**Primitive Schema Handling** decides what happens when a *Record Reader* is also configured.
+`Record Reader if configured` — the default — parses the payload with the reader, which is what a `STRING`
+topic carrying JSON or CSV text wants. `Single-field record` always wraps the value instead.
+
+The choice is a property rather than an inference from whether a reader is set, because the reader is
+*also* the fallback for topics with no schema: configuring one for that reason should not silently change
+how primitive topics are read. Under the default, a `STRING` topic carrying plain text with a
+`JsonTreeReader` configured sends every message to `parse.failure`, since the reader cannot parse it and
+the single-field record is not reachable — `Single-field record` is the setting for that flow.
 
 Publishing to a primitive topic requires a record with **exactly one field**, whose value is coerced to
 the topic's type. A record with several fields has no unambiguous mapping onto a single value, so it is
--- a/README.md
+++ b/README.md
@@ -141,8 +141,28 @@ with.
 
 A *Record Reader* may still be configured under `Topic Schema`, and messages the strategy cannot decode
 fall back to it — a topic with no schema at all, or one whose schema has no record shape. Without a
-reader to fall back to, those messages go to `parse.failure`. Primitive and `KeyValue` schemas are not
-yet decoded this way.
+reader to fall back to, those messages go to `parse.failure`. `KeyValue` schemas are not yet decoded this
+way.
+
+### Topics with a primitive schema
+
+A topic whose schema is a primitive — `STRING`, `BOOLEAN`, or one of the numeric types — carries one
+value per message and has no fields, so `Topic Schema` gives each message a record of a single field
+named by **Primitive Value Field** (`value` by default). It becomes a column name downstream, so it is
+worth setting to something meaningful.
+
+**A configured *Record Reader* takes precedence on these topics.** A `STRING` topic carrying JSON text is
+a common shape, and those flows want the payload parsed into records rather than wrapped in one string
+field — so with a reader configured the payload is parsed, and the single-field record is what happens
+when there is no reader to parse with.
+
+Publishing to a primitive topic requires a record with **exactly one field**, whose value is coerced to
+the topic's type. A record with several fields has no unambiguous mapping onto a single value, so it is
+routed to `failure` rather than guessing which field was meant.
+
+`BYTES` is not treated as a primitive schema. Pulsar reports a topic with *no* schema as `BYTES` with an
+empty definition, so the two are indistinguishable, and treating it as primitive would capture every
+schema-less topic. The date and time schemas are not supported yet either.
 
 ## Publishing to topics that have a schema
 
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarRecord.java
@@ -113,6 +113,30 @@ public class ConsumePulsarRecord extends AbstractPulsarConsumerProcessor<Generic
             .defaultValue(SCHEMA_FROM_RECORD_READER.getValue())
             .build();
 
+    static final AllowableValue PRIMITIVE_VIA_READER = new AllowableValue("Record Reader if configured",
+            "Record Reader if configured",
+            "Parse the payload with the Record Reader when one is configured, and wrap it in a single field "
+            + "only when there is none. Suits a STRING topic carrying JSON or CSV text.");
+
+    static final AllowableValue PRIMITIVE_AS_RECORD = new AllowableValue("Single-field record",
+            "Single-field record",
+            "Always wrap the value in a single field, whether a Record Reader is configured or not. Suits a "
+            + "topic whose values are genuinely scalar, and leaves the reader free to serve as the fallback "
+            + "for topics that have no schema.");
+
+    public static final PropertyDescriptor PRIMITIVE_SCHEMA_HANDLING = new PropertyDescriptor.Builder()
+            .name("PRIMITIVE_SCHEMA_HANDLING")
+            .displayName("Primitive Schema Handling")
+            .description("What to do with a topic whose schema is a primitive. Only used by the 'Topic "
+                    + "Schema' strategy. The default defers to the Record Reader when one is configured, "
+                    + "which is what a STRING topic carrying JSON text wants; choose 'Single-field record' "
+                    + "when the values really are scalar, so that configuring a reader as the schema-less "
+                    + "fallback does not change how primitive topics are read.")
+            .required(false)
+            .allowableValues(PRIMITIVE_VIA_READER, PRIMITIVE_AS_RECORD)
+            .defaultValue(PRIMITIVE_VIA_READER.getValue())
+            .build();
+
     public static final PropertyDescriptor PRIMITIVE_VALUE_FIELD = new PropertyDescriptor.Builder()
             .name("PRIMITIVE_VALUE_FIELD")
             .displayName("Primitive Value Field")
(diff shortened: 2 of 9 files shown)


david-streamlio:
@jselvareef this one is directly from your point 4 on #184 — a review would be very welcome before it goes in, and I will wait for it this time.

The precedence rule is the part I would most like you to check. A configured Record Reader wins on a primitive topic, so a STRING topic carrying JSON text is parsed into records rather than wrapped; with no reader, the single-field record is what you get. That means **the same topic behaves differently depending on whether a reader is configured**, which I have documented as deliberate — but if you think an explicit property would be less surprising than inferring it from whether a reader is set, now is the moment to say so.

Two other things worth your eye:

1. **`BYTES` is excluded.** An integration test caught that `AUTO_PRODUCE_BYTES` reports a schema-*less* topic as `BYTES` with an empty definition, so a registered BYTES schema is indistinguishable from none, and supporting it broke the schema-less path this bundle's own publishers produce. I am fairly confident that is right, but you know the history better — was a genuinely registered BYTES schema ever a case anyone relied on?

2. **Publishing requires a single-field record**, and a multi-field one goes to `failure`. The alternative would be taking a named field, which is another property. Failing seemed better than guessing, but it is a judgement call about what users will hit.

Deferred deliberately: the date/time schemas, for the same reason AVRO logical types are still uncovered in #34.

#190 (KeyValue) is next and I will put it up for review the same way.

jselvareef (commented):
Reviewed the code and ran everything locally (podman): 299 unit tests and the five relevant IT classes green — `PrimitiveSchemaRoundTripIT` 7/7 plus the inherited `ConsumePulsarTopicSchemaIT`, `NiFiSchemaRoundTripIT`, `ConsumePulsarRecordIT`, `NiFiRoundTripIT`. Encoding through Pulsar's own `Schema.STRING`/`INT32`/… and checking every encode against the client's decoder is the right way to do this. Answers to the three questions, then two minor notes.

### 1. The precedence rule

The rule does what I asked for on #184 — a STRING topic carrying JSON keeps parsing into records — but it has a blind spot the tests do not show, because none of them combines a reader with a primitive topic whose payload is *not* parseable. Probe: *Topic Schema*, a `JsonTreeReader` configured (the README's own advice, since the reader is also the schema-less fallback), a STRING topic carrying plain text:

```
2 messages "hello world" / "second line"  →  success=0  parse_failure=1  (both messages)
```

So once a reader is configured for any reason, the single-field record becomes unreachable for that processor: a plain-string topic can only fail, and the user's only way out is to remove the reader — which also removes the schema-less fallback. Inferring the behaviour from "is a reader set" ties two unrelated decisions together. I would make it explicit: a property such as **Primitive Schema Handling** with `Record Reader if configured` (the default — today's behaviour, nothing changes for anyone) and `Single-field record`. Deterministic per processor, no new default, and the two ITs that pin the rule today become the two values.

### 2. `BYTES` excluded — yes

Agreed, and the history supports it. A `Schema.BYTES` producer registers nothing — #184 showed exactly that (`Schema not found` after publishing with the default schema) — so the only way to get a *registered* BYTES schema is an explicit admin upload, and nothing in this bundle's past ever did that: before #171 the publishers used `Schema.BYTES` and left every topic they created schema-less, and the consumers have always subscribed with `AUTO_CONSUME`. Treating BYTES as "no schema" matches what the broker does with it.

### 3. Multi-field records on publish — failing is right, with one cheap escape hatch

Refusing to guess is correct as the default. The symmetric answer to the consumer's *Primitive Value Field* would be the same property on `PublishPulsarRecord`: when set, that field is the value and the rest of the record is ignored; when not set, the record must have exactly one field, as now. It gives a flow with an existing multi-field record a path that does not need an extra processor, and it mirrors the consumer so a round trip can name the field once. Not blocking — the guard's error message already names the fields, which is the part that matters.

### Minor

- `PrimitiveTopicSchema.encode` for STRING uses `DataTypeUtils.toString(value, (String) null)`, so a nested record or map value is published as its `toString()` rather than rejected; probably fine, but worth a test if it is meant to be allowed.
- `decodePrimitive` allocates a `HashMap` per message for one entry; `Collections.singletonMap` would do, since `MapRecord` is not written to afterwards.

Everything else reads right: the publish guard is gated by *Message Schema Strategy* = `Topic Schema`, so existing flows on the default strategy are untouched; `NumberFormatException` is caught into the `IOException` path; and the README states the precedence rule plainly.

david-streamlio:
Thanks — all three points acted on, plus both minors. Pushed as `ecb0d72`, with the #194/#195 fixes in `735114e`.

**Precedence rule.** You were right and I have made it a property: **Primitive Schema Handling**, `Record Reader if configured` (default, so nothing changes for anyone) or `Single-field record`. Your probe is now an IT — a STRING topic of plain text with a `JsonTreeReader` configured, which under the default fails and under the new value wraps — plus one asserting the default still defers to the reader. The reasoning went in the README too: the reader is *also* the schema-less fallback, so configuring it for that reason should not silently decide this.

**#194 and #195**, since #192 carries the override and the decoder: both fixed here.

- `customValidate` now seeds from `super.customValidate()`, with tests asserting the no-topic and 2-second-timeout configurations are invalid.
- The decoder is created per `consumeMessages()` call so nothing is shared, **and** its cache is one immutable snapshot behind a volatile, so the class is safe even if a later change shares it again. Your two-thread probe is now a test: 20,000 decodes each over two schemas, asserting no record comes back with the other's shape.

Reverting both: `Tests run: 15, Failures: 3` → `0` after.

**BYTES** — thank you for the history, that settles it.

**Multi-field publish escape hatch** — I have not done it. It is a good idea and symmetric with the consumer, but it is a new property on a second processor and this PR is already carrying two bug fixes; I would rather it were its own change. Say if you would rather see it here.

**Minors:** the `HashMap` is now `Collections.singletonMap`. On `toString`, I went further than a test — a record, map, array or iterable is now **refused** for a STRING topic rather than published as its `toString()`, since `MapRecord[{id=1}]` on a topic looks like success and is not. Scalars still render.

Full suite: 304 unit tests, 60 ITs.

jselvareef:
Verified on `ecb0d72` (podman): 305 unit tests, `PrimitiveSchemaRoundTripIT` 9/9 including the plain-text case under both values of *Primitive Schema Handling*, and the #194/#195 fixes hold against my original probes — no-topic and 2-second-timeout configurations invalid again, and 0 wrong records in 40,000 for both the AVRO/AVRO and the STRING/INT32 two-thread runs. Refusing structured values on a STRING topic instead of publishing their `toString()` is the better call.

On the publish-side escape hatch: agreed, its own change. Nothing here should wait for it. Looks good to me.

PR #193, opened 2026-08-30
feat: map KeyValue-schema topics to key and value record fields (#190)

### LINKED ISSUE #190: Support KeyValue schema topics in the record processors

david-streamlio:
Closes #190. **Stacked on #192** (`feat/189-primitive-schemas`) — it reuses the primitive mapping for key sides, so it is based on that branch rather than `main`. Retarget to `main` once #192 merges.

## The change

A `KEY_VALUE` topic carries two schemas and an encoding that says where the key lives:

- **`INLINE`** — key and value both length-prefixed into the payload.
- **`SEPARATED`** — key in the message's key metadata, value in the payload. This is what makes a topic compactable by key, and is the more common shape.

Neither was recognised before: records went through the Record Writer and the broker accepted whatever happened to match.

`Topic Schema` now maps each message to a record with two fields, named by **KeyValue Key Field** and **KeyValue Value Field**. Each side keeps the shape its own schema describes — a `STRING` key becomes a string field, an `AVRO` value a nested record — so the shape downstream is what the topic actually describes, rather than two nested records by convention.

The INLINE layout is Pulsar's, so `KeyValue.encode` / `KeyValue.decode` build and split it rather than our own arithmetic over length prefixes.

## The `msg.key` collision, resolved by refusing

This was the design question in the issue. On a `SEPARATED` topic the schema's key field **becomes** the message key — so *Message Key Field* cannot also own it, and the two would overwrite each other.

It refuses: the FlowFile goes to `failure` with a message naming both properties. Unlike the `CustomPartition` rejection in #182, this **cannot be caught at validation** — the topic's schema is only known at publish time — so it is a runtime failure by necessity, not by choice.

## Tests

`KeyValueSchemaRoundTripIT`, against a real broker:

- **e2e NiFi→NiFi round trips** on `SEPARATED` and on `INLINE`
- a JSON value side as well as AVRO
- the configurable field names
- *Message Key Field* refused on a SEPARATED topic
- a record missing either side refused

**299 unit tests and 64 integration tests green.**

## Not covered

- A KeyValue schema whose key or value is **itself** a KeyValue.
- The consume-side interaction with *Mapped FlowFile Attributes*: the mapped `msg.key` still gives the raw encoded key rather than the decoded one. Arguably it should surface the decoded key, but that changes an existing attribute's meaning and belongs in its own change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,22 @@ routed to `failure` rather than guessing which field was meant.
 empty definition, so the two are indistinguishable, and treating it as primitive would capture every
 schema-less topic. The date and time schemas are not supported yet either.
 
+### Topics with a KeyValue schema
+
+A `KEY_VALUE` topic carries two schemas — one for the key, one for the value — and an encoding that says
+where the key is written. `INLINE` length-prefixes both into the payload; `SEPARATED` puts the key in the
+message's key metadata and only the value in the payload, which is what makes a topic compactable by key.
+Both are supported and behave the same to a flow.
+
+`Topic Schema` gives each message a record with two fields, named by **KeyValue Key Field** and
+**KeyValue Value Field** (`key` and `value` by default). Each side keeps the shape its own schema
+describes: a `STRING` key becomes a string field, an `AVRO` value becomes a nested record.
+
+Publishing needs a record with both of those fields. **On a `SEPARATED` topic the key field becomes the
+message key, so *Message Key Field* must not also be set** — the two would overwrite each other, and the
+FlowFile is routed to `failure` rather than silently letting one win. The topic's schema is not known
+until publish time, so this cannot be caught when the processor is configured.
+
 ## Publishing to topics that have a schema
 
 `PublishPulsar` and `PublishPulsarRecord` create their producers with
--- a/README.md
+++ b/README.md
@@ -151,10 +151,15 @@ value per message and has no fields, so `Topic Schema` gives each message a reco
 named by **Primitive Value Field** (`value` by default). It becomes a column name downstream, so it is
 worth setting to something meaningful.
 
-**A configured *Record Reader* takes precedence on these topics.** A `STRING` topic carrying JSON text is
-a common shape, and those flows want the payload parsed into records rather than wrapped in one string
-field — so with a reader configured the payload is parsed, and the single-field record is what happens
-when there is no reader to parse with.
+**Primitive Schema Handling** decides what happens when a *Record Reader* is also configured.
+`Record Reader if configured` — the default — parses the payload with the reader, which is what a `STRING`
+topic carrying JSON or CSV text wants. `Single-field record` always wraps the value instead.
+
+The choice is a property rather than an inference from whether a reader is set, because the reader is
+*also* the fallback for topics with no schema: configuring one for that reason should not silently change
+how primitive topics are read. Under the default, a `STRING` topic carrying plain text with a
+`JsonTreeReader` configured sends every message to `parse.failure`, since the reader cannot parse it and
+the single-field record is not reachable — `Single-field record` is the setting for that flow.
 
 Publishing to a primitive topic requires a record with **exactly one field**, whose value is coerced to
 the topic's type. A record with several fields has no unambiguous mapping onto a single value, so it is
--- a/README.md
+++ b/README.md
@@ -141,8 +141,28 @@ with.
 
 A *Record Reader* may still be configured under `Topic Schema`, and messages the strategy cannot decode
 fall back to it — a topic with no schema at all, or one whose schema has no record shape. Without a
-reader to fall back to, those messages go to `parse.failure`. Primitive and `KeyValue` schemas are not
-yet decoded this way.
+reader to fall back to, those messages go to `parse.failure`. `KeyValue` schemas are not yet decoded this
+way.
+
+### Topics with a primitive schema
+
+A topic whose schema is a primitive — `STRING`, `BOOLEAN`, or one of the numeric types — carries one
+value per message and has no fields, so `Topic Schema` gives each message a record of a single field
+named by **Primitive Value Field** (`value` by default). It becomes a column name downstream, so it is
+worth setting to something meaningful.
+
+**A configured *Record Reader* takes precedence on these topics.** A `STRING` topic carrying JSON text is
+a common shape, and those flows want the payload parsed into records rather than wrapped in one string
+field — so with a reader configured the payload is parsed, and the single-field record is what happens
+when there is no reader to parse with.
+
+Publishing to a primitive topic requires a record with **exactly one field**, whose value is coerced to
+the topic's type. A record with several fields has no unambiguous mapping onto a single value, so it is
+routed to `failure` rather than guessing which field was meant.
+
+`BYTES` is not treated as a primitive schema. Pulsar reports a topic with *no* schema as `BYTES` with an
+empty definition, so the two are indistinguishable, and treating it as primitive would capture every
+schema-less topic. The date and time schemas are not supported yet either.
 
 ## Publishing to topics that have a schema
 
--- a/README.md
+++ b/README.md
@@ -181,9 +181,15 @@ Both are supported and behave the same to a flow.
 describes: a `STRING` key becomes a string field, an `AVRO` value becomes a nested record.
 
 Publishing needs a record with both of those fields. **On a `SEPARATED` topic the key field becomes the
-message key, so *Message Key Field* must not also be set** — the two would overwrite each other, and the
-FlowFile is routed to `failure` rather than silently letting one win. The topic's schema is not known
-until publish time, so this cannot be caught when the processor is configured.
+message key, so *Message Key Field* must not name a different field** — the two would overwrite each
+other, and the FlowFile is routed to `failure` rather than silently letting one win. Naming the *same*
+field is allowed: that asks for what the schema already guarantees. The topic's schema is not known until
+publish time, so this cannot be caught when the processor is configured.
+
+Because the schema's key becomes the message key on a `SEPARATED` topic, it is also the routing key — the
+same key lands on the same partition, and the topic is compactable by it, without configuring anything.
+On an `INLINE` topic the key metadata is unused by the schema, so *Message Key Field* still works there as
+the routing key.
 
 ## Publishing to topics that have a schema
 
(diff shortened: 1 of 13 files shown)


david-streamlio:
@jselvareef review welcome on this one too — I will wait for it rather than merging, same as #192.

Note it is **stacked on #192**, so review that one first; this branch is based on it because a KeyValue key side is usually a primitive and reuses that mapping. I will retarget to `main` once #192 goes in.

The design decision I would most like challenged is the **`msg.key` collision**. On a SEPARATED topic the schema's key field becomes the message key, so *Message Key Field* cannot also own it. I made that a runtime failure routing the FlowFile to `failure`, on the reasoning that silently letting one overwrite the other is worse. Unlike the `CustomPartition` rejection in #182 it cannot be a validation error, because the topic's schema is not known until publish time. If you would rather it ignored *Message Key Field* with a warning, that is a defensible alternative and easy to change.

Second thing worth your view: on the consume side, *Mapped FlowFile Attributes* still surfaces `msg.key` as the **raw encoded** key on a SEPARATED topic, which is not meaningful once a key schema is involved. I left that alone deliberately — changing it alters an existing attribute's meaning — but it is arguably now wrong, and I would rather hear whether you think it should be a follow-up issue.

Also flagged in the PR: a KeyValue schema whose key or value is itself a KeyValue is not handled.

jselvareef (commented):
Reviewed the code and ran it locally (podman): 299 unit tests, and `KeyValueSchemaRoundTripIT` 6/6 plus the inherited `PrimitiveSchemaRoundTripIT`, `ConsumePulsarTopicSchemaIT` and `NiFiSchemaRoundTripIT` — all green. Using `KeyValue.encode`/`decode` for the INLINE layout and `keyBytes()` for SEPARATED is the right mirror of what Pulsar's own `KeyValueSchema` does. The two questions first, then what I found.

### 1. The `msg.key` collision — refusing is right, with one exemption

Silently letting either side win would change partition routing and compaction without a trace, so failing the FlowFile is the better default, and you are right that it cannot be a validation error. One refinement: when *Message Key Field* names the **same** field as *KeyValue Key Field* there is no conflict — the user is asking for exactly what the schema already guarantees — and failing there would surprise anyone who set both for clarity. Refuse only when they differ. Two things worth a sentence in the README while you are there: on a SEPARATED topic the schema key also becomes the routing key (same key → same partition, for free), and on an INLINE topic *Message Key Field* still works as the routing key, since the key metadata is free — the code already behaves that way.

### 2. The mapped `msg.key` on SEPARATED topics — define it now, it costs nothing

`msg.getKey()` returns the **base64 text** of a key set with `keyBytes()`, so under *Mapped FlowFile Attributes* a SEPARATED topic yields base64 even for a STRING key, which nobody can use. Since the KeyValue-plus-*Topic Schema* combination is new in this PR, there is no existing flow whose attribute meaning would change — the only observers of that attribute today see base64. So the rule can be set here rather than in a follow-up: for a primitive key schema, the decoded value as a string; for a struct key, leave the attribute alone (or omit it). If you prefer to keep this PR to the record mapping, a follow-up issue is fine, but it does not have to wait for compatibility reasons. (Unchanged and worth remembering: a mapped `msg.key` is part of the record-set key, so mapping it puts every distinct key in its own FlowFile.)

### Findings

**A nested record on a JSON side fails to publish.** `TopicSchemaRecordDecoder.encodeStruct` builds a `LinkedHashMap` of `record.getValue(field)` and hands it to Jackson, so a field that is itself a record arrives as a NiFi `MapRecord` object:

```
schema Order{id:int, customer:Customer{name:string}}, JSON side
→ InvalidDefinitionException: Java 8 optional type `java.util.Optional<java.lang.String>` not supported by default
  (through reference chain: LinkedHashMap["customer"] -> MapRecord["schema"] -> SimpleRecordSchema["schemaFormat"])
```

The FlowFile goes to `failure` for any JSON value side with a nested record; `KeyValueSchemaRoundTripIT` does not see it because its value schema is flat (`Reading{reading:int32}`). The top-level JSON path from #178 already does this right — convert through `AvroTypeUtil.createAvroRecord` and write the Avro record as plain JSON — so the fix is to share that method rather than keep a second JSON encoding.

**Struct sides are parsed on every message.** `decodeSide` calls `dataTypeOf(side)`, which runs `Schema.Parser().parse` and `AvroTypeUtil.createSchema` per message, then parses the Avro schema a second time for the datum reader; `encodeStruct` parses per record. `parse()` already caches per definition — the side schemas (Avro and NiFi) belong in that cache. The primitive and top-level struct paths cache; this one is the outlier.

**`encodingOf()` has no caller**, and if it ever gets one before `decode()`, it hands `parse()` the still-null `cachedKeyField`/`cachedValueField`, which become `RecordField` names. Remove it, or make it take the field names.

**`keyValueDecoder` is another processor-level field with an unsynchronized cache**, so #195 applies to it as well — whichever fix lands there should cover both decoders. The `PublisherLease` copy is fine: a lease serves one FlowFile at a time.

**No unit tests.** The INLINE split/encode symmetry and both refusal rules are testable without a broker through Pulsar's own `KeyValue.encode`/`decode`, and a nested-record case at that level is where the JSON finding would have surfaced.

Nested KeyValue not being handled is fine as declared.

david-streamlio:
All of it acted on — pushed as `bd76c7c`. The nested-record finding was the important one, and it turned out to have a mirror image.

**The JSON nesting defect.** Confirmed and fixed by sharing `writeAsJson` from #178 rather than keeping a second encoding, exactly as you suggested. Writing the nested unit test you asked for then found the same bug on the way *back*: JSON decoding built a `MapRecord` straight from Jackson's map, and **neither** `MapRecord` constructor converts a nested object — so a nested field came back as a raw `LinkedHashMap`. Both directions now go through `DataTypeUtils.toRecord`.

Worth flagging: **the decode half is latent in #185 as merged**, not new in this PR. Any JSON topic with a nested record returns raw Maps today on `main`. It is fixed here, but if #192/#193 stall it is worth pulling out separately.

I also took the first fix that did not work as a signal to stop guessing — `checkTypes=true` looked like the answer and was not — and checked the API instead.

**Side schemas** are now parsed once per definition. **`encodingOf()`** removed. **The KeyValue decoder** is created per batch like the record decoder, so #195 does not apply to it either.

**The `msg.key` same-field exemption** — taken. Only a genuine disagreement fails now; naming the same field publishes. That flipped my existing test, which set `MESSAGE_KEY_FIELD=key` against a default *KeyValue Key Field* of `key` — so it is now two tests, one for each case. Both README sentences you asked for are in, including that INLINE leaves *Message Key Field* usable as the routing key.

**Unit tests** — `KeyValueTopicSchemaTest`, 9 of them: both encodings round tripped through Pulsar's own `KeyValue.encode`/`decode`, the nested-record case that found the defect above, both refusal rules, a SEPARATED message with no key, and a malformed INLINE frame.

**Mapped `msg.key` on SEPARATED topics** — not done here, and I would like your view first. Your compatibility argument is right that nothing depends on the base64 today. But the decoded key is only meaningful for a primitive key schema, so the rule becomes "decoded string for a primitive key, attribute absent for a struct key" — and an attribute that appears or vanishes depending on the key schema's type feels like something to decide deliberately rather than fold in here. Happy to do it in this PR if you think that is over-thinking it; otherwise I will file it.

Full suite: 313 unit tests, 67 ITs. Still stacked on #192.

jselvareef:
Verified on `bd76c7c` (podman): 316 unit tests with `KeyValueTopicSchemaTest` 9/9, `KeyValueSchemaRoundTripIT` 7/7, and the nested case round-trips now — `{"id":7,"customer":{"name":"Ana"}}` out, `MapRecord` back with the name intact. Same-field exemption, per-batch decoder, side-schema cache and `encodingOf` removal all as described.

**On the latent decode half in `main`:** I checked how far it reaches. `TopicSchemaRecordDecoder` on `main` does return the nested field as a `LinkedHashMap`, but `JsonRecordSetWriter` converts a `Map` to a record on write for a RECORD-typed field, so the FlowFile content comes out right (`[{"id":7,"customer":{"name":"Ana"}}]`). It only bites code that reads the field through the Record API expecting a `Record`, which nothing in this bundle does. So it is worth fixing — you have — but not worth pulling out ahead of #192/#193.

**On the mapped `msg.key` for SEPARATED topics:** file it. Your appear-or-vanish objection is right, and I would answer it by never vanishing: the attribute is always the decoded key — the value's string form for a primitive key, its JSON rendering (through `writeAsJson`, now that it is shared) for a struct key. Never base64. That keeps one rule for the attribute regardless of the key schema, and it does not belong in a PR that is already carrying the mapping.

Nothing else from me; looks good.

Issue #207, opened 2026-08-30
Avro decoding ignores __AVRO_READ_OFFSET__, so Kafka Connect / Debezium topics decode from the wrong offset


david-streamlio:
Found while confirming the struct-key use case for #198. Independent of that issue, and with a wider blast radius: this affects the **value** side as well as the key.

## The problem

Pulsar carries a schema property `__AVRO_READ_OFFSET__` telling a consumer how many bytes to skip before the Avro body begins. Pulsar's own reader honours it:

```java
// pulsar-client/.../generic/GenericAvroReader.java:70-71
if (schema.getObjectProp(GenericAvroSchema.OFFSET_PROP) != null) {
    this.offset = Integer.parseInt(schema.getObjectProp(GenericAvroSchema.OFFSET_PROP).toString());
```

**This bundle never reads that property.** `grep -rn 'AVRO_READ_OFFSET' ` over the repo returns nothing, and both decoders start at byte zero:

- `nifi-pulsar-processors/.../utils/KeyValueTopicSchema.java:255-256`
- `nifi-pulsar-processors/.../utils/TopicSchemaRecordDecoder.java:159-163`

```java
new GenericDatumReader<GenericRecord>(avroSchema)
        .read(null, DecoderFactory.get().binaryDecoder(data, null));
```

## Who sets it, and to what

Any topic produced through Pulsar's Kafka Connect adaptor — which includes **every Debezium source connector**, since `DebeziumSource extends KafkaConnectSource`. From `kafka-connect-adaptor/.../schema/KafkaSchemaWrappedSchema.java`:

```java
boolean isJsonConverter = converter instanceof JsonConverter;
props.put(GenericAvroSchema.OFFSET_PROP, isJsonConverter ? "0" : "5");
...
.type(isJsonConverter ? SchemaType.JSON : SchemaType.AVRO)
```

`5` is the Confluent wire format: one magic byte plus a four-byte schema id ahead of the Avro body. So on such a topic our decoder feeds the magic byte and schema id to Avro as if they were record data.

## Expected impact

A `GenericDatumReader` reading from the wrong offset either throws or returns wrong values, depending on what the leading bytes happen to decode as. Both outcomes are bad and the second is worse: `ConsumePulsarRecord` catches decode failures and falls back, so a throw degrades quietly, and a successful-but-wrong decode is silent corruption.

[...]

PR #211, opened 2026-08-30
fix: honour __AVRO_READ_OFFSET__ when decoding Avro (#207)

david-streamlio:
Addresses #207.

## Confirmed before writing anything

- `__AVRO_READ_OFFSET__` is real and Pulsar honours it — `GenericAvroReader` reads it and starts its decoder there.
- This bundle never referenced it; both decode sites started at byte zero.

## Why it is silent

Against the unfixed decoder, end to end through a real broker — five framed messages published, all five consumed:

```
[{"id":"","reading":0},{"id":"","reading":0},{"id":"","reading":0},{"id":"","reading":0},{"id":"","reading":0}]
```

Every field empty or zero, **routed to `success`, no parse failures**. Avro's binary encoding is permissive enough that the wrong bytes decode to a plausible record rather than failing, and `ConsumePulsarRecord` treats a decode failure as a fallback — so both paths are quiet. The tests therefore assert decoded field values, not absence of a throw; a test checking only for parse failures would pass against the bug.

## One correction to the fix proposed on the issue

The issue suggests reading the property from `SchemaInfo.getProperties()`. Pulsar reads it from the **parsed Avro schema**:

```
113: ldc  #128  // String __AVRO_READ_OFFSET__
115: invokevirtual  Schema.getObjectProp:(Ljava/lang/String;)Ljava/lang/Object;
```

Following that suggestion would have produced a fix that compiled, passed a test built on the same assumption, and did nothing in production — the same failure mode as the bug. This reads the Avro schema property first, matching Pulsar, and falls back to `SchemaInfo` properties, which costs nothing and covers a producer that records it there instead.

Absent, unparseable or negative all mean offset 0 — today's behaviour, and what every topic not framed this way depends on.

## The fix

- Both decode sites: the record path, and **each side of a KeyValue topic independently**, since a KeyValue schema frames key and value separately.
- A message shorter than its declared offset fails rather than decoding whatever the arithmetic produces — message and schema disagree, which is worth surfacing.

## Tests

**`TopicSchemaRecordDecoderTest`** — the fast regression, so it runs in CI's Build & Test job and not only behind Docker.

**`AvroReadOffsetIT`** — against a real broker, because the unit test cannot prove the part that involves one: that a schema carrying a non-standard property survives registration and returns on `AUTO_CONSUME`'s reader schema intact. A broker that dropped it would leave the unit test passing and production broken. Three tests: a framed topic decoded past its preamble; a topic with no offset property still read from byte zero; and the property read back from the broker, so a broker that dropped it fails loudly instead of quietly reducing the first test to a no-op.

Fail-first, with the offset lookup neutralised: **1 of 3 fails**. The other two are controls and correctly do not depend on the fix.

**324 unit tests and 70 integration tests green**, rebased on `64755dd` so #208's `decodeKeyAsText` / `toJsonText` and this coexist.

## Scope note

The issue asked for a **CDC-backed** integration test with a real Debezium source. David decided to take this smaller broker-backed IT for now instead. So what remains unverified is narrow but real: that a live Debezium source emits exactly this framing. The preamble here is written by hand from the Kafka Connect adaptor's documented behaviour — magic byte plus four-byte schema id — rather than observed from a running CDC stack. Worth keeping on #207 rather than closing it silently.

### DIFF
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/utils/AvroReadOffset.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/utils/AvroReadOffset.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.processors.pulsar.utils;
+
+import java.io.IOException;
+
+import org.apache.pulsar.common.schema.SchemaInfo;
+
+/**
+ * How many bytes precede the Avro body in a message (#207).
+ * <p>
+ * A schema may carry {@code __AVRO_READ_OFFSET__}, and Pulsar's own {@code GenericAvroReader} honours it:
+ * it reads the property off the parsed Avro schema and starts its decoder there. The Kafka Connect adaptor
+ * sets it to {@code 5} for Avro — the Confluent wire format's magic byte plus a four-byte schema id — so
+ * every topic fed by a Debezium source carries a five-byte preamble ahead of the record.
+ * <p>
+ * Decoding from byte zero reads that preamble as record data. It rarely fails loudly: Avro's binary
+ * encoding is permissive enough that the wrong bytes usually decode to a plausible-looking record, so the
+ * symptom is wrong values rather than an exception.
+ * <p>
+ * The property is read from the <em>parsed Avro schema</em> first, which is where Pulsar looks and
+ * therefore the authoritative source. {@link SchemaInfo#getProperties()} is consulted as a fallback, since
+ * it costs nothing and a producer that records the offset there rather than in the schema document would
+ * otherwise be silently mishandled. Absent, unparseable or negative all mean zero — today's behaviour, and
+ * the default every topic that is not framed this way depends on.
+ */
+public final class AvroReadOffset {
+
+    static final String PROPERTY = "__AVRO_READ_OFFSET__";
+
+    private AvroReadOffset() {
+    }
+
+    /**
+     * @param avroSchema the parsed schema, which is where Pulsar reads the property from
+     * @param schemaInfo the topic's schema info, consulted only if the schema itself does not carry it
+     * @return the number of bytes to skip, never negative
+     */
+    public static int of(final org.apache.avro.Schema avroSchema, final SchemaInfo schemaInfo) {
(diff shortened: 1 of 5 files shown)


jselvareef:
Ran the branch (`046d66d`) here against a broker (Testcontainers, `apachepulsar/pulsar:4.2.4`): `TopicSchemaRecordDecoderTest` 11/11, `AvroReadOffsetIT` 3/3. The fix itself reads right — taking the property off the parsed Avro schema first is the correct call, and the broker-preservation test is the one that makes the other two mean something. Two things I would raise before merging, both found by pointing the *other* paths at the same framed topic the IT builds.

### 1. The Record Reader path still decodes the preamble as data

The PR fixes the two Topic Schema decode sites. The default strategy — a Record Reader, which for an AVRO topic the README says to configure as an `AvroReader` on `${avro.schema}` — is untouched, and on a framed topic it produces exactly the symptom #207 describes. Same framed topic as the IT (`__AVRO_READ_OFFSET__ = 5`, three framed records), `ConsumePulsarRecord` with the default strategy and an `AvroReader` on `${avro.schema}`:

```
success = [{"id":"","reading":0},{"id":"","reading":0},{"id":"","reading":0},{"id":"","reading":0},
           {"id":"","reading":-58},{"id":"","reading":0},{"id":"","reading":0}]
parse_failure = 1
```

Seven "records" out of three messages, routed to `success`; the `avro.schema` attribute carries `__AVRO_READ_OFFSET__` faithfully, but NiFi's `AvroReader` has no idea what it means. So after this PR the outcome on a framed topic depends on which strategy the flow happens to use — silent garbage on the documented default, correct records on *Topic Schema* — with nothing in the docs saying so.

Two ways out. The processor already has the message's `SchemaInfo` when it hands the payload to the reader, so it could apply the same `AvroReadOffset.of(...)` there and give the reader `data` from the offset on — one call, and both strategies agree. Failing that, the README and `additionalDetails` should say plainly that a topic declaring a read offset must be read with *Topic Schema*.

### 2. Publishing with Topic Schema to a framed topic fails on every message, with a message that names the wrong culprit

`PublishPulsarRecord`, *Message Schema Strategy* = `Topic Schema`, against the same topic:

```
Unable to Publish Pulsar Records
org.apache.pulsar.shade.org.apache.avro.AvroRuntimeException: Malformed data. Length is negative: -56
    at ...GenericAvroReader.read(GenericAvroReader.java:86)
    at ...AbstractStructSchema.decode(AbstractStructSchema.java:66)
```

Every FlowFile to `failure`. The producer is `AUTO_PRODUCE_BYTES`, so the client validates the payload by decoding it with its own `GenericAvroReader` — which honours the offset. The bare Avro the lease writes starts five bytes too early for the topic's own contract, and the message the user sees is Avro complaining about a string length, not "this topic expects a five-byte preamble this publisher does not write". Loud rather than silent, so a lesser problem than #207 — but the same property, and the opposite direction of the same interop story.

If the aim is to interoperate with topics fed by the Kafka Connect adaptor, the lease could frame the encoded record with `offset` bytes when the topic schema declares one (Pulsar's readers only skip them, so zero bytes satisfy every consumer that honours the property). If that is out of scope, refusing with a clear message when the topic schema carries the property would at least name the cause. Either way this deserves an issue of its own rather than a silent "not covered"; I can open it if you agree.

### Smaller

- No README / `additionalDetails` change in the PR. The consumers' *Consuming from topics that have a schema* section is where a user hitting this will look; a paragraph on framed topics — what the property is, which strategy honours it, what to expect from `ConsumePulsar`, which passes the bytes through as they are — would close the loop.
- `AvroReadOffset.check` fails the decode when the message is shorter than the declared offset, which is right, but the decode path logs that at `debug` and the message then follows the fallback route, so the one case that clearly says "schema and message disagree" is the hardest to see. `warn` for that one seems worth it.

Nothing here argues against the fix as written; the two decode sites are correct, and the fail-first note (1 of 3 with the lookup neutralised) matches what I would expect.

PR #217, opened 2026-09-02
feat: finish Topics Pattern, and add Read Compacted and Subscription Mode (#187)

### LINKED ISSUE #187: Add missing Consumer configuration properties
## Method

The same two passes as #186, over `AbstractPulsarConsumerProcessor`:

1. **Declared but not wired** — every `PropertyDescriptor` in the consumer hierarchy, checked for a `getProperty(NAME)` read anywhere in `src/main`.
2. **Not declared at all** — every field of `ConsumerConfigurationData` (pulsar-client 4.2.4) against what we expose.

## Part 1 — declared but never read

**None.** Every consumer property reaches the client.

That is not luck: `getConsumerBuilder()` sets each one through a `ConsumerBuilder` method, so an unused property is dead code the compiler and reader can both see. The producer instead passes a `Map` to `loadConf()`, where a missing entry looks like nothing at all — which is how #180 happened. Worth keeping the consumer on builder calls for that reason alone.

## Part 2 — supported by the client, not exposed

Currently wired: `topicNames`, `topicsPattern`, `subscriptionName`, `subscriptionInitialPosition`, `subscriptionType`, `autoUpdatePartitions` (+ interval), `ackTimeoutMillis`, `autoAckOldestChunkedMessageOnQueueFull`, `expireTimeOfIncompleteChunkedMessageMillis`, `maxPendingChunkedMessage`, `consumerName`, `priorityLevel`, `receiverQueueSize`, `replicateSubscriptionState`.

**Worth adding, roughly in order of how often they matter:**

| Field | Why |
|---|---|
| `deadLetterPolicy` | The largest gap. A message that fails to parse goes to `parse.failure` and is acknowledged, so NiFi handles poison messages — but there is no way to route them to a DLQ topic, and no `maxRedeliverCount` ceiling. Anyone used to Pulsar's DLQ semantics will expect this. |
| `negativeAckRedeliveryDelayMicros`, `negativeAckRedeliveryBackoff` | We never call `negativeAcknowledge()` at all; unacknowledged messages wait out `ackTimeout` instead. Negative acking failures would redeliver far sooner, and the delay wants to be configurable. |
| `readCompacted` | Reading a compacted topic — the latest value per key — is a standard pattern and currently impossible. |
| `subscriptionMode` | `Durable` vs `NonDurable`. A non-durable subscription leaves no cursor behind, which is what you want for tailing. |
| `regexSubscriptionMode` | We already expose **Topics Pattern**, but not whether it matches persistent, non-persistent, or all topics. Defaults to persistent-only, so a pattern silently misses non-persistent topics. |
| `patternAutoDiscoveryPeriod` | Also pairs with the existing Topics Pattern: how quickly a newly created matching topic is picked up. |
| `keySharedPolicy` | We offer `Key_Shared` as a subscription type but no policy — sticky vs auto-split ranges, and whether out-of-order delivery is allowed. |
| `acknowledgementsGroupTimeMicros`, `maxAcknowledgmentGroupSize` | Acknowledgement batching. Directly affects broker load at high throughput. |
| `retryEnable` | Pulsar's built-in retry topic, the usual companion to a dead letter policy. |
| `maxTotalReceiverQueueSizeAcrossPartitions` | The partitioned counterpart to Receiver Queue Size. |
| `startPaused` | Subscribe without consuming until the flow is ready. |
| `batchIndexAckEnabled` | Acknowledge individual messages inside a batch instead of whole batches. |
| `cryptoKeyReader`, `cryptoFailureAction` | Decryption, matching the producer side in #186. Needs a controller service. |
| `subscriptionProperties` | Subscription metadata, visible in topic stats. |
| `autoScaledReceiverQueueSizeEnabled`, `resetIncludeHead`, `tickDurationMillis`, `negativeAckPrecisionBitCnt`, `poolMessages` | Tuning knobs; lower value, listed for completeness. |

**Deliberately excluded:** `messageListener`, `consumerEventListener`, `messageListenerExecutor`, `payloadProcessor`, `batchReceivePolicy`, `messageCrypto`, `decryptFailListener`, `topicConfigurations` — needs user code, or is managed by the processor's own polling loop.

## Suggested split

`deadLetterPolicy` and the negative-acknowledgement settings belong together and are the highest value — they are also the ones that change the processor's behaviour, not just its configuration, so they deserve their own PR and integration test. `readCompacted`, `subscriptionMode`, `regexSubscriptionMode`, and `patternAutoDiscoveryPeriod` are small additions and the last two finish a feature already half-exposed.

Related: #186 (the producer half of the same audit).

david-streamlio:
The remaining behaviour tranche of #187 — four consumer settings the Pulsar client has always
supported and no flow could reach. All four default to the client's own default, so no existing
flow changes behaviour.

| Property | The gap it closes |
|---|---|
| *Topics Pattern Match Mode* | a pattern **silently missed non-persistent topics** |
| *Topics Pattern Discovery Interval* | no control over how fast a new matching topic is picked up |
| *Subscription Mode* | `NonDurable` (tailing, no backlog accrual) was unreachable |
| *Read Compacted* | the latest-value-per-key view was unreachable |

### Topics Pattern was half-exposed

*Topics Pattern* has shipped for years, but a pattern matches **persistent topics only** by
default and there was no way to say otherwise — so a pattern that plainly matches a
non-persistent topic consumed nothing, with nothing anywhere explaining why. There was also no
control over the discovery sweep, which is the worst-case delay before a newly created matching
topic is read. These two finish the feature.

### Read Compacted needed a validator

The client refuses it at subscribe time:

> `Read compacted can only be used with exclusive or failover persistent subscriptions`

`Shared` is this processor's **default** subscription type, so the most likely configuration
would validate cleanly and then fail on every schedule. Now rejected up front, as
`CustomPartition` and the dead letter policy are.

### The two single-consumer constraints are mirror images

Worth stating plainly, because nothing else in the codebase does:

- a **compacted read** needs a *single active* consumer → `Exclusive` / `Failover`
- a **dead letter policy** needs *competing* consumers → `Shared` / `Key_Shared`

**No subscription type satisfies both**, so they can never be enabled together. A test pins this
by flipping the subscription type and asserting invalid either way — otherwise it is discovered
by someone who configured both and cannot see why neither works.

### Two decisions worth reviewing

**The pattern properties are set unconditionally**, not guarded on a pattern being in use. The
client reads them only when it has a pattern, and mirroring that condition here would be a second
place to keep in step with the topics/pattern choice above it.

**They are accepted with a topic list rather than rejected.** This deliberately differs from the
dead-letter-topic-without-a-count rejection: a dead letter topic that can never receive a message
misleads someone into waiting for it, whereas a match mode on a subscription doing no matching
misleads nobody — and rejecting it would make both awkward to leave configured while switching
between a list and a pattern.

### Verification

- **359 unit tests** pass (349 before, +10).
- **Integration tests against a real broker**, covering what a mock cannot — a compacted read is
  the broker serving a different view, and what a pattern matches is decided broker-side:
  - a pattern with `NonPersistentOnly` consumes from a non-persistent topic;
  - a compacted read returns only the latest value per key.
- **Both guards mutation-tested:**
  - dropping the Read Compacted validator fails **3 of 10** unit cases;
  - dropping `subscriptionTopicsMode` fails the IT with *"Timed out waiting for the pattern to
    discover the non-persistent topic"* — the reported bug, restated.

The compaction test triggers compaction and waits for the admin CLI to report
`Compaction was a success`. That wait is load-bearing: an uncompacted topic serves its normal
backlog, so without it the assertions would pass while measuring nothing. (My first attempt
matched the wrong string and would have done exactly that.)

### Documentation

All four property descriptions written — those are NiFi's generated docs. Plus a *Choosing what
to consume* section on both consumers' `additionalDetails.html` and in the README, including the
mirror-image constraint.

Closes #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,33 @@ is what to use when clients in other languages also write the topic.
 > either raise *Max Pending Messages*, set it to `0` to restore the previous unbounded behaviour, or
 > enable *Block if Message Queue Full* so sends wait instead of failing.
 
+## Choosing what to consume
+
+*Topics* and *Topics Pattern* are alternatives; exactly one must be set.
+
+A pattern matches **persistent topics only** by default, which is why a pattern that plainly
+matches a non-persistent topic can appear to do nothing. *Topics Pattern Match Mode* changes
+that — `PersistentOnly` (the default and the previous behaviour), `NonPersistentOnly`, or
+`AllTopics`. *Topics Pattern Discovery Interval* is how often the client re-evaluates the
+pattern, and so the worst-case delay before a newly created matching topic is read. Both are
+ignored when *Topics* is used.
+
+*Subscription Mode* decides whether the broker keeps a cursor. `Durable` (the default) survives
+a restart and resumes where it left off. `NonDurable` leaves no cursor: the subscription exists
+only while the consumer is connected, which is what tailing wants, and it accumulates no backlog
+on the broker while the flow is stopped.
+
+*Read Compacted* reads the compacted view of a topic — the latest value per key — instead of its
+backlog. Messages without a key are not delivered at all, and the topic must actually have
+compaction running for there to be a compacted view; without it the subscription reads the normal
+backlog.
+
+> **The two single-consumer constraints are mirror images.** A compacted read needs a *single
+> active consumer*, so Pulsar permits it only on `Exclusive` and `Failover`. A dead letter policy
+> needs *competing* consumers, so Pulsar builds one only for `Shared` and `Key_Shared`. No
+> subscription type satisfies both, and the processor rejects each on the wrong type at validation
+> rather than letting the client fail at subscribe time.
+
 ## Producer behaviour
 
 *Send Timeout* bounds how long a single send may take. A message the broker has not acknowledged
--- a/README.md
+++ b/README.md
@@ -208,8 +208,11 @@ with exclusive or failover **persistent** subscriptions"*.
 - **A single active consumer** — so `Exclusive` or `Failover`.
 - **The persistent domain** — only a persistent topic has a compacted view. A `non-persistent://`
   topic named literally in *Topics* is rejected at validation, as is a *Topics Pattern* whose *Match
-  Mode* admits non-persistent topics. A topic supplied by an expression cannot be checked until the
-  expression resolves, so that case is warned about at startup instead. That second case matters because the client cannot catch it: with a pattern its topic list
+  Mode* admits non-persistent topics. A topic supplied by an expression is only checked if the
+  expression resolves at startup — an environment variable or a system property does, and a
+  non-persistent topic from one of those is warned about when the processor starts. A topic taken
+  from FlowFile attributes does not resolve at startup, so neither validation nor the warning can
+  see it and the client is what refuses the subscription. That second case matters because the client cannot catch it: with a pattern its topic list
   is empty, so its own domain check passes vacuously and any non-persistent topic the pattern matches
   is served as a live stream — the flow would read a full stream while its configuration says it is
   reading the latest value per key, with nothing reporting it. A `non-persistent://` prefix on the
--- a/README.md
+++ b/README.md
@@ -184,12 +184,18 @@ matches a non-persistent topic can appear to do nothing. *Topics Pattern Match M
 that — `PersistentOnly` (the default and the previous behaviour), `NonPersistentOnly`, or
 `AllTopics`. *Topics Pattern Discovery Interval* is how often the client re-evaluates the
 pattern, and so the worst-case delay before a newly created matching topic is read. Both are
-ignored when *Topics* is used.
+ignored when *Topics* is used — with one exception: the discovery interval is still bounded to a
+whole number of seconds within `int` range whichever is set, because the client checks that bound
+when it builds any consumer, not only a pattern one.
 
 *Subscription Mode* decides whether the broker keeps a cursor. `Durable` (the default) survives
 a restart and resumes where it left off. `NonDurable` leaves no cursor: the subscription exists
 only while the consumer is connected, which is what tailing wants, and it accumulates no backlog
-on the broker while the flow is stopped.
+on the broker while the flow is stopped. What it gives up is delivery across restarts — with no
+cursor there is no resume point, so a `NonDurable` subscription set to *Subscription Initial
+Position* `Earliest` re-reads the topic from the beginning every time it is scheduled, and every
+time a consumer is evicted from the cache. The processor warns when it starts in that state.
+Worth knowing because the *Read Compacted* guidance below sends you to `Earliest`.
 
 *Read Compacted* reads the compacted view of a topic — the latest value per key — instead of its
 backlog. Messages without a key are not delivered at all, and the topic must actually have
--- a/README.md
+++ b/README.md
@@ -196,11 +196,28 @@ backlog. Messages without a key are not delivered at all, and the topic must act
 compaction running for there to be a compacted view; without it the subscription reads the normal
 backlog.
 
+*Read Compacted* has two requirements, and the client states both: *"Read compacted can only be used
+with exclusive or failover **persistent** subscriptions"*.
+
+- **A single active consumer** — so `Exclusive` or `Failover`.
+- **The persistent domain** — only a persistent topic has a compacted view. A `non-persistent://`
+  topic is rejected at validation, as is a *Topics Pattern* whose *Match Mode* admits non-persistent
+  topics. That second case matters because the client cannot catch it: with a pattern its topic list
+  is empty, so its own domain check passes vacuously and any non-persistent topic the pattern matches
+  is served as a live stream — the flow would read a full stream while its configuration says it is
+  reading the latest value per key, with nothing reporting it.
+
 > **The two single-consumer constraints are mirror images.** A compacted read needs a *single
 > active consumer*, so Pulsar permits it only on `Exclusive` and `Failover`. A dead letter policy
 > needs *competing* consumers, so Pulsar builds one only for `Shared` and `Key_Shared`. No
 > subscription type satisfies both, and the processor rejects each on the wrong type at validation
 > rather than letting the client fail at subscribe time.
+
+> **Concurrent Tasks and non-Shared subscriptions.** Because *Read Compacted* requires `Exclusive` or
+> `Failover`, flows using it are on a subscription type where *Concurrent Tasks* > 1 is not currently
+> safe: one task's cumulative acknowledgement can acknowledge messages another task still holds. See
+> [#223](https://github.com/david-streamlio/pulsar-nifi-bundle/issues/223). Run these flows with a
+> single task until that is resolved.
 
 ## Producer behaviour
 
--- a/README.md
+++ b/README.md
@@ -201,8 +201,9 @@ with exclusive or failover **persistent** subscriptions"*.
 
 - **A single active consumer** — so `Exclusive` or `Failover`.
 - **The persistent domain** — only a persistent topic has a compacted view. A `non-persistent://`
-  topic is rejected at validation, as is a *Topics Pattern* whose *Match Mode* admits non-persistent
-  topics. That second case matters because the client cannot catch it: with a pattern its topic list
+  topic named literally in *Topics* is rejected at validation, as is a *Topics Pattern* whose *Match
+  Mode* admits non-persistent topics. A topic supplied by an expression cannot be checked until the
+  expression resolves, so that case is warned about at startup instead. That second case matters because the client cannot catch it: with a pattern its topic list
   is empty, so its own domain check passes vacuously and any non-persistent topic the pattern matches
   is served as a live stream — the flow would read a full stream while its configuration says it is
   reading the latest value per key, with nothing reporting it. A `non-persistent://` prefix on the
--- a/README.md
+++ b/README.md
@@ -205,7 +205,14 @@ with exclusive or failover **persistent** subscriptions"*.
   topics. That second case matters because the client cannot catch it: with a pattern its topic list
   is empty, so its own domain check passes vacuously and any non-persistent topic the pattern matches
   is served as a live stream — the flow would read a full stream while its configuration says it is
-  reading the latest value per key, with nothing reporting it.
+  reading the latest value per key, with nothing reporting it. A `non-persistent://` prefix on the
+  *pattern itself* is not rejected, because it is inert: the client strips the scheme from the pattern
+  and from every candidate topic, so the domain comes from *Match Mode* alone.
+
+> **Set *Subscription Initial Position* to `Earliest`.** The compacted view is the history of the
+> topic — the latest value for each key seen so far. A new subscription left at the default of
+> `Latest` starts at the tail, so a compacted read delivers **nothing at all** until a new message
+> arrives, which looks identical to a broken flow.
 
 > **The two single-consumer constraints are mirror images.** A compacted read needs a *single
 > active consumer*, so Pulsar permits it only on `Exclusive` and `Failover`. A dead letter policy
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
@@ -55,7 +55,9 @@ import org.apache.pulsar.client.api.Message;
 import org.apache.pulsar.client.api.PulsarClientException;
 import org.apache.pulsar.client.api.Schema;
 import org.apache.pulsar.common.schema.SchemaInfo;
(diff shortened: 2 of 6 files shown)


jselvareef:
One gap in the *Read Compacted* validator, found while running the branch against a broker.

The validator checks the subscription type, but the client's own check (`PulsarClientImpl.subscribeAsync`, 4.2.4) has a second leg: every topic must be in the `persistent://` domain. So this configuration validates in NiFi and fails on every schedule:

- *Topic Names* = `non-persistent://public/default/x`, *Subscription Type* = `Exclusive`, *Read Compacted* = `true`

Reproduced on `58e5c14` with `ConsumePulsar` (Testcontainers `apachepulsar/pulsar:4.2.4`): `MockProcessContext.isValid()` → `true`; the first trigger throws

```
ProcessException: org.apache.pulsar.client.api.PulsarClientException$InvalidConfigurationException:
Read compacted can only be used with exclusive or failover persistent subscriptions
    at org.apache.nifi.processors.pulsar.AbstractPulsarConsumerProcessor.getConsumer(AbstractPulsarConsumerProcessor.java:671)
```

— the same message the PR quotes for the `Shared` case, reached by the other half of the client's condition. Since a topic list is a known set of names at validation time, rejecting any `non-persistent://` entry there when *Read Compacted* is set closes it the way the subscription-type rule does.

For a *Topics Pattern* the client cannot check (the topic set is empty at subscribe), so a pattern with *Topics Pattern Match Mode* = `NonPersistentOnly` or `AllTopics` plus *Read Compacted* gets past both validators; what the broker then does with a compacted read on a non-persistent topic I have not exercised, so I would only flag that combination in the property description rather than reject it.

david-streamlio:
You were right, and I owe you an apology for the duplicated effort — I found this independently two
days after you posted it, because I went to a fresh review pass without reading the comments on my
own PR first. Entirely avoidable.

Fixed in `8f38178`, and the topic-list half is exactly as you described: any `non-persistent://`
entry in *Topics* is rejected when *Read Compacted* is set. An unqualified name stays valid, since it
resolves to the persistent domain, and an expression is left to the client, which refuses it once
resolved. Mutation-tested — reverting to the subscription-type-only version fails 4 of the 16 unit
cases.

**On the pattern case I went further than you suggested, and I am no longer sure I was right.**

You said you would flag the combination in the property description rather than reject it, since you
had not exercised what the broker does. I decompiled that path: with a pattern, `conf.getTopicNames()`
is empty, so the client's `allMatch(persistent)` guard passes vacuously, and
`NonPersistentTopic.internalSubscribe` does not carry `readCompacted` at all. So a non-persistent
match is subscribed and served as a live stream — the flow reads a full stream while its
configuration says it is reading the latest value per key, and nothing reports it.

That is worse than an error, which is what pushed me to reject: *Topics Pattern Match Mode* must be
`PersistentOnly` when *Read Compacted* is set, plus a rejection for a pattern in the non-persistent
domain.

The problem with that, and the reason I am raising it rather than closing it: **`AllTopics` +
*Read Compacted* is not wholly broken.** The persistent matches *are* served compacted, correctly.
Only the non-persistent ones are silently wrong. So a hard rejection removes a capability that half
works, and someone consuming a pattern that mostly matches persistent topics loses it entirely.

Three ways to go, and I do not have a strong preference:

1. **Reject** (what is on the branch). Safe, and no flow can silently read the wrong thing. Costs the
   partially-working `AllTopics` case.
2. **Your original suggestion** — allow it, describe the caveat on the property. Keeps the capability;
   relies on someone reading the description to avoid a silent wrong read, which is the failure mode
   this bundle has generally chosen to design out.
3. **Allow, but warn at `@OnScheduled`** naming both properties — the hook you used for the delay
   inversion in #220. Keeps the capability and puts the caveat somewhere it is actually seen, once
   per start rather than never.

I lean 3 now, mostly because you established the pattern in #220 and it fits this shape better than
it fits that one: there the configuration was merely suboptimal, here it is silently wrong for a
subset of topics, which is worth a line in the log every start. Happy to switch the branch to it.

One more thing, unrelated to your comment but found in the same review pass and worth your attention
since it touches what you are doing in #221: **cumulative acknowledgement loses messages on a
non-Shared subscription with Concurrent Tasks > 1** — one task's `acknowledgeCumulative` acknowledges
messages another task still holds, and a write failure in that task then loses them. Pre-existing,
not from this branch, filed as #223. It matters here only because *Read Compacted* requires
`Exclusive`/`Failover`, so anyone using it is on the affected path by construction; the branch now
documents that and points at the issue.

jselvareef:
No apology needed — the fix in `8f38178` is the one I would have written, and the pattern half is the part I had not run, so you have taken it further than my comment did.

On the three options: I would **keep the rejection**, and I think the premise that it costs a half-working capability does not quite hold.

With *Read Compacted* set, `AllTopics` gives the persistent matches compacted — correctly — plus the non-persistent matches as a live stream, silently. `PersistentOnly` gives *exactly the same persistent matches, compacted*, and nothing else. So the correctly-working half of `AllTopics` is already available in full through `PersistentOnly`; the only thing the rejection removes is the half that is wrong. Nobody loses a topic that was being read correctly, and the message on the branch already tells them which setting to use (*"Topics Pattern Match Mode must be PersistentOnly, not AllTopics"*). Someone who genuinely wants the non-persistent matches too wants them uncompacted — the broker has nothing else to offer for them — and that is a second processor without *Read Compacted*, not a mode on this one.

That is also what separates it from #220's case. There the inverted delay was *suboptimal* and might be deliberate (a long backoff), so a warning was the honest level. Here the configuration is *wrong for a subset of topics with no error*, the correct configuration is one value away, and the rejection points at it. A warning would keep the silent wrong read available for the sake of a capability that `PersistentOnly` already provides.

`NonPersistentOnly` + *Read Compacted* is wrong for every match, so that rejection stands on its own.

Two small things from the later commits, both agreeing:

- `defc7f0` dropping the pattern-domain check is right — I had the prefix down as inert too, having read `TopicsPatternFactory`, which is why my original suggestion was only about the topic list. Pinning the inert-prefix case with a test so the rejection cannot come back is the useful part.
- The *Read Compacted* + `Latest` warning fits the warn-at-schedule shape exactly: the flow does deliver on a live topic, so it is unusual rather than wrong. Same reasoning as above, opposite conclusion, which I think is the consistent line.

On #223: that one is in #224, with a probe showing the loss needs only one task — a nacked message is acknowledged away by the next batch's cumulative acknowledgement — so it is on the *Read Compacted* path by construction regardless of *Concurrent Tasks*. Dropping the concurrency IT here was right; the interleaving it would have hit is the one #224's IT pins on purpose.

david-streamlio:
Cross-PR note, not a review of this branch — raised here so the two PRs converge deliberately rather than by accident.

## Your whole-seconds validator is the one that should win

#228 (which closes #225, the issue the comment at `AbstractPulsarConsumerProcessor.customValidate` defers to) fixes the two sibling time properties, and it answers the fractional-value question the *other* way: it rejects only sub-second values and **warns** about a fractional one like `90500 millis` instead of rejecting it. So as things stand, two adjacent time properties on the same processor would treat the same input differently.

I've asked #228 to adopt your approach rather than asking you to loosen this one. Reasoning, since it contradicts the compatibility argument in your own comment here:

- Sub-second values are **already** broken on `Auto Update Partition Interval` — `ConsumerConfigurationData.setAutoUpdatePartitionsIntervalSeconds` does `checkArgument(interval > 0, "interval needs to be > 0")`, so those flows throw on every trigger today. Rejecting them at validation loses nothing that runs.
- That leaves only fractional-but-≥1s values as the compatibility exposure, on a property whose default is `1 min` — it had to be typed deliberately.
- The break is a processor that goes invalid on the canvas with a message that explains itself, versus a flow that silently runs at 90 s while its configuration says 90.5. The loud failure is better.

Also worth noting for the docs on both sides: NiFi time-period values always carry their unit (`TIME_PERIOD_VALIDATOR`: "Must be of format `<duration> <TimeUnit>` … nanos, millis, secs, mins, hrs, days"), so there is no ambiguity to resolve about what a bare number means. The only question per property is whether the duration the user expressed is representable in the granularity the client stores it in — whole seconds for `patternAutoDiscoveryPeriod` and `autoUpdatePartitionsIntervalSeconds`, milliseconds for `expireTimeOfIncompleteChunkedMessageMillis`. Framed that way the three rules fall out without needing a judgement call.

## Two things this branch shares with the gap I flagged on #228

**1. No upper bound, and `patternAutoDiscoveryPeriod` takes an `int`.** The validator here checks `millis < 1000L || millis % 1000L != 0L` but nothing above, while the builder call passes `.asTimePeriod(TimeUnit.SECONDS).intValue()`. Verified arithmetic:

| Configured | Seconds | `.intValue()` | Result |
|---|---|---|---|
| `30000 days` | 2,592,000,000 | −1,702,967,296 | `checkArgument(interval >= 0)` throws |
| `10000 weeks` | 6,048,000,000 | 1,753,032,704 | silently runs as ≈55 y, not ≈191 y |

Adding `|| TimeUnit.MILLISECONDS.toSeconds(millis) > Integer.MAX_VALUE` to the existing condition closes both.

**2. The validation is gated on `TOPICS_PATTERN.isSet()`, but the builder call is not.** That asymmetry is deliberate and the comment explains why — but combined with the overflow above it has a consequence I don't think was intended: a **topic-list** flow that happens to carry `30000 days` in this property skips validation entirely, then hits `patternAutoDiscoveryPeriod(-1702967296, SECONDS)` and throws `interval needs to be >= 0` on every trigger — failing a flow over a property that has no effect on it, which is exactly what the gating was written to avoid. Either bound the value ungated, or guard the builder call the way the validation is guarded.

Nothing here needs to block this PR; #2 is the one I'd want closed before merge.

Separately, on merge order: this branch, #222 and #228 all modify `README.md`, and #222 and #228 both create `docs/release-notes/2.11.0.2.md`. All three read mergeable against `main` today, but only until the first one lands.

david-streamlio:
Correction to my note above — **items 1 and 2 there are one bug, not two, and one check fixes both.** I presented item 2 as an independent flaw in the gating and offered "either bound the value ungated, or guard the builder call" as alternatives. That framing would send you into more surgery than this needs.

**Your gating premise is right where it matters.** `PatternMultiTopicsConsumerImpl` really is the only thing that reads the value, so the property really is inert at runtime for a topic-list flow. The one thing the comment doesn't account for is that `ConsumerBuilderImpl.patternAutoDiscoveryPeriod(int, TimeUnit)` runs `checkArgument(interval >= 0)` at **build time**, for every consumer regardless of topics-vs-pattern. Gated validation and an unguarded builder call therefore only diverge where a validated value is one the builder rejects — and since `TIME_PERIOD_VALIDATOR` already guarantees a non-negative duration, the *only* way to reach the builder negative is an `int` overflow. Which is item 1.

So item 2 is item 1's symptom:

| Configured | Seconds | `.intValue()` | Builder |
|---|---|---|---|
| `500 millis` | 0 | 0 | passes (`>= 0`) |
| `10000 weeks` | 6,048,000,000 | 1,753,032,704 | passes — silently ≈55 y, not ≈191 y |
| `30000 days` | 2,592,000,000 | **−1,702,967,296** | **throws** |

Worth being explicit that this means **`aBadDiscoveryIntervalIsIgnoredWithATopicList` is legitimately green**, and I did not mean to imply otherwise: `500 millis` truncates to `0`, `0 >= 0` passes, and `Math.max(1, …)` makes it a 1-second sweep nothing reads. That test asserts something true.

## The whole fix

One unconditional check ahead of the existing gated block. Nothing else moves:

```java
// The builder's precondition (interval >= 0) runs for every consumer, pattern or not, and an int
// overflow is the only way a non-negative duration reaches it negative - so this one cannot be
// gated on TOPICS_PATTERN even though the value itself is inert without a pattern.
final long millis = validationContext.getProperty(PATTERN_AUTO_DISCOVERY_PERIOD)
        .asTimePeriod(TimeUnit.MILLISECONDS);
if (TimeUnit.MILLISECONDS.toSeconds(millis) > Integer.MAX_VALUE) {
    results.add(new ValidationResult.Builder().valid(false)
        .subject(PATTERN_AUTO_DISCOVERY_PERIOD.getDisplayName())
        .explanation("the client keeps this as an int number of seconds, so it cannot exceed "
            + Integer.MAX_VALUE + " seconds (about 68 years)").build());
}
```

Explicitly retracting both suggestions from my earlier note:

- **Don't guard the builder call.** Your comment already argues against it, and once the range is enforced there is nothing left to protect against — it would mean breaking the fluent chain for no gain.
- **Don't un-gate the whole-seconds check.** That check is about whether a sweep interval is *meaningful*, which genuinely only applies with a pattern. Un-gating it would fail a topic-list flow carrying `1500 millis` over an inert property — the thing your gating exists to prevent — and would turn the test above red for the wrong reason.

Two tests would pin it: a topic-list flow with `30000 days` → `assertNotValid` (today it validates and then throws on every trigger), and the boundary, `2147483647 sec` valid / `2147483648 sec` not.

Everything else in my earlier note stands, including the merge-order overlap on `README.md` and the request on #228 to adopt your whole-seconds rule — where this same bound closes the identical gap on *Auto Update Partition Interval*, so the two properties end up with the same rule shape.

PR #220, opened 2026-09-05
fix: default the negative ack redelivery delay under the ack timeout, and warn when it is not (#218)

### LINKED ISSUE #218: Default Negative Acknowledgment Redelivery Delay (1 min) makes a write failure stall 60 s, twice the Acknowledgment Timeout it was meant to beat
With the processor defaults, the negative acknowledgement added in #216 makes a message the processor could not write come back **later** than before, not sooner: about 60 s on `main` against about 30 s on `v2.10.0.1`, measured against a real broker with the same test.

### Where

`AbstractPulsarConsumerProcessor.rollbackAndRedeliver()` calls `consumer.negativeAcknowledge(message)` for every message of the rolled-back batch. In the Pulsar client that call does two things (`ConsumerImpl.negativeAcknowledge(Message)`, checked on `pulsar-client` 4.2.4):

1. adds the message to the `NegativeAcksTracker`, which redelivers it after `negativeAckRedeliveryDelay`;
2. **removes it from the `UnAckedMessageTracker`** — the client's comment reads "Ensure the message is not redelivered for ack-timeout, since we did receive an 'ack'".

So once a message is nacked, *Acknowledgment Timeout* no longer applies to it at all. The only thing that redelivers it is *Negative Acknowledgment Redelivery Delay*, and that property defaults to `1 min` while *Acknowledgment Timeout* defaults to `30 sec` (floor `10 sec`). In the default configuration the "ceiling" the release notes describe is not a ceiling: the nacked message waits the full minute.

### Reproduction

`ConsumePulsarDeadLetterIT` with the two timing properties **left at their defaults** instead of `1 sec` / `10 sec`: subscribe (Shared), publish one message, run one trigger with a session whose `write()` throws, then poll with a healthy session and time the redelivery. Same source run on both trees, Testcontainers `apachepulsar/pulsar:4.2.4`:

| Tree | Redelivered after |
|---|---|
| `main` (`00a6eb8`, with #216) | **60.4 s** |
| `64755dd` (`v2.10.0.1`, rollback only) | **30.6 s** |

The existing IT does not see this because it sets the delay to `1 sec` and the timeout to `10 sec` ("so the test does not wait on the one-minute default"), and then asserts the redelivery lands inside the timeout. With those values it does; with the defaults the relation is inverted.

### Impact

- Every write failure (full or read-only content repository, disk fault) now stalls the batch for 60 s by default instead of 30 s — the opposite of what #216 set out to do, and a behaviour change for every existing flow, none of which has the new property set.
- The release notes say the broker "redelivers them at once" and, further down, that the gap "defaults to one minute". Both sentences are in the same document; only the second is what a default flow gets.
- With a dead letter policy the effect compounds: *Max Redelivery Count* is counted per redelivery, so reaching the dead letter topic takes `count × 1 min` by default.

### Options

- Default *Negative Acknowledgment Redelivery Delay* below the *Acknowledgment Timeout* floor — the IT's own `1 sec` is one candidate; anything under 10 s restores the promised behaviour for a default flow. Pulsar's one-minute default is tuned for an application nacking a poison message, not for an infrastructure failure the processor wants retried promptly.
- Add a validation rule that the delay is not longer than *Acknowledgment Timeout*, so a tuned flow cannot reintroduce the inversion, and say in the property description that a nacked message is governed by this delay alone.
- Run the timing assertion in `ConsumePulsarDeadLetterIT` at the defaults, since that is the configuration the claim is made about.

Refs #187, #216.

jselvareef:
Closes #218.

A negatively acknowledged message is redelivered by *Negative Acknowledgment Redelivery Delay* **alone**: `ConsumerImpl.negativeAcknowledge` (checked on `pulsar-client` 4.2.4) adds the message to the `NegativeAcksTracker` and removes it from the `UnAckedMessageTracker` — "since we did receive an 'ack'". So from the nack onward *Acknowledgment Timeout* no longer applies to that message. With the delay defaulting to `1 min` and the timeout to `30 sec`, a flow that never set the property waited **60 s** after a write failure on `main` where `v2.10.0.1` waited 30 s (measured in #218) — the opposite of what #216 set out to do.

> **Revised after review** (second commit): the default is `5 sec` rather than `1 sec`, a delay longer than the timeout is warned about at scheduling rather than rejected at validation, the IT's write-failure detection is specific and monotonic, and there are release notes. Details in the review thread.

### What changed

- **`AbstractPulsarConsumerProcessor`** — the delay defaults to `5 sec`, under the `10 sec` floor the validator puts on *Acknowledgment Timeout*, so the invariant "a nacked message comes back sooner than an unacknowledged one would have" holds for every valid flow — without collapsing the *Max Redelivery Count* budget the way `1 sec` would have (60× faster than `2.11.0`; now 12×, called out in the docs). A configured delay longer than *Acknowledgment Timeout* stays **valid** — a flow that set one on `2.11.0` must still start after upgrading, and a long delay is a legitimate poison-message backoff — but the processor logs a warning naming both properties when it is scheduled. NiFi's `ValidationResult` has no warning level, so `@OnScheduled` is where that lives. The property description now says that the delay is the only thing governing a nacked message and that each redelivery counts against *Max Redelivery Count*.
- **`ConsumePulsarDeadLetterIT`** — the redelivery timing test runs at the **default** delay, since the default is what the claim is about. It subscribes *before* publishing: with the message published first, the initialising pass (which runs with a healthy session) could consume it, the failing pass then received nothing, and the test passed with no negative acknowledgement ever happening — that is how it stayed green with a one-minute default. The failing pass is repeated until a **new** `"Unable to write the received messages"` error has been logged (count snapshotted before the loop), so an unrelated error from an earlier pass cannot end the loop early. The dead letter test sets the delay explicitly to `1 sec`, since its pace depends on the delay and it is not the test of the default.
- **`ConsumePulsarNegativeAckDelayTest`** (new, 7 cases) — pins the default-below-floor invariant, that shorter/equal/longer all validate, that longer produces exactly one warning on scheduling naming both properties, that within-timeout produces none, and that the comparison is on durations rather than strings.
- **Docs** — README *Failure handling and redelivery* states the new default, the warning and the effect on *Max Redelivery Count*, with a *Behaviour change since `2.11.0`* callout; both consumers' `additionalDetails.html` say the same. New `docs/release-notes/2.11.0.1.md` (the next connector revision under VERSIONING.md) carries the behaviour change and an *Upgrading* section; `2.11.0.md` keeps its text as the record of that release, with a one-line pointer at the change.

### Fails first

Unit tests on `main` (`00a6eb8`), first version of the test class — 3 of the 6 cases fail:

```
Tests run: 6, Failures: 3, Errors: 0, Skipped: 0 -- in ConsumePulsarNegativeAckDelayTest
  theDefaultDelayIsShorterThanTheShortestAllowedAcknowledgmentTimeout:
    the default Negative Acknowledgment Redelivery Delay (60000 ms) is not shorter than the 10 s floor of
    Acknowledgment Timeout, so a write failure would stall longer than the timeout it was meant to beat
  aDelayLongerThanTheAcknowledgmentTimeoutIsRejected:
    Processor appears to be valid but expected it to be invalid ==> expected: <false> but was: <true>
  theComparisonIsOnDurationsNotOnTheWrittenValues:
    Processor appears to be valid but expected it to be invalid ==> expected: <false> but was: <true>
```

(The two "rejected" cases are now "warns when scheduled"; on `main` they fail the same way, with no warning logged.)

The reordered IT on `main`, at the default delay — both tests fail, and the second one shows the dead letter policy is hit by the same default (redelivery count advances once a minute):

```
Tests run: 2, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 106.6 s -- in ConsumePulsarDeadLetterIT
  aMessageThatCouldNotBeWrittenIsRedeliveredWithoutWaitingOutTheAckTimeout -- 31.11 s
    java.lang.AssertionError: Timed out after 30s waiting for: the negatively acknowledged message to be redelivered
  aRepeatedlyUndeliverableMessageEndsUpOnTheDeadLetterTopic -- 60.08 s
    java.lang.AssertionError: the poison message never reached the dead letter topic
```

With the fix as it stands (`5 sec` default):

```
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 32.60 s -- in ConsumePulsarDeadLetterIT
```

### Verification

- `mvn test` on the current head: **356/356** (349 + 7).
- First commit: `mvn -B -ntp verify -Dgpg.skip=true -pl nifi-pulsar-processors -am` (the CI command) against a real broker via Testcontainers on rootless podman: **71 integration tests, 68 pass in the full run**. The 3 errors are `java.io.IOException: Broken pipe` from podman's exec endpoint inside `execInContainer` (`pulsar-admin` creating partitioned topics), thrown before any processor runs — an environment problem here, not the tests. Re-running those classes on their own, `ConsumePulsarRecordIT`, `NiFiRoundTripIT` and `PublishPulsarRoutingModeIT` all pass.
- Second commit: `ConsumePulsarDeadLetterIT` 2/2 against a real broker; the other ITs were not re-run since the change touches only the default value, `@OnScheduled` and this IT.

### Not done

- Nothing about *which* rollback sites nack, or about `parse_failure`, changes.
- The comparison uses the configured values directly; neither property supports Expression Language, so there is nothing to evaluate at validation time.
- The warning is logged once per scheduling, not on a timer; a flow that runs for weeks with the inverted configuration has one line in the log to find.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,8 +97,19 @@ The third row is the one to know about. When the processor cannot write a messag
 **negatively acknowledges** the message, which asks the broker to redeliver it now. Without that
 the message is merely unacknowledged, and the broker cannot tell a consumer that has failed from
 one that is still working: it waits out *Acknowledgment Timeout*, thirty seconds by default and
-never less than ten. Set *Negative Acknowledgment Redelivery Delay* to control how soon; it
-defaults to Pulsar's own one minute.
+never less than ten. *Negative Acknowledgment Redelivery Delay* controls how soon the redelivery
+comes; it defaults to one second, and cannot be set longer than *Acknowledgment Timeout*. The
+reason for that rule: once a message is negatively acknowledged the client stops tracking it for
+the timeout, so the delay is the **only** thing that redelivers it — a delay longer than the timeout
+would make a write failure wait longer than a plain rollback did.
+
+> **Behaviour change since `2.11.0`:** in `2.11.0` the delay defaulted to Pulsar's own one minute,
+> so with *Acknowledgment Timeout* at its 30-second default a message the processor could not write
+> came back after **60 s — twice as long as before negative acknowledgement existed**, not sooner.
+> The default is now one second, so a flow that never set the property redelivers after a write
+> failure in about a second instead of a minute. A flow that had set the property to a value above
+> its *Acknowledgment Timeout* is now invalid and has to lower one or raise the other. To keep the
+> old gap on purpose, set the delay explicitly to a value no longer than the timeout.
 
 A message routed to `parse_failure` is **not** redelivered. It was delivered and handled — the
 flow has its bytes and can route them anywhere, including back to a Pulsar topic — so nacking it
--- a/README.md
+++ b/README.md
@@ -17,7 +17,8 @@ The bundle version tracks the NiFi platform version it is built for; each releas
 line targets one Pulsar client major. See [VERSIONING.md](VERSIONING.md) for the
 full scheme, branching model, and release process.
 
-Release notes live in [`docs/release-notes/`](docs/release-notes/). `2.11.0` is a
+Release notes live in [`docs/release-notes/`](docs/release-notes/); the notes for the next
+revision on the current line accumulate in [`2.11.0.1`](docs/release-notes/2.11.0.1.md). `2.11.0` is a
 platform bump — see [its notes](docs/release-notes/2.11.0.md). If you are coming from
 `2.9.0` or earlier, read [the `2.10.0` notes](docs/release-notes/2.10.0.md) too: that
 release carries several behaviour changes.
@@ -98,18 +99,23 @@ The third row is the one to know about. When the processor cannot write a messag
 the message is merely unacknowledged, and the broker cannot tell a consumer that has failed from
 one that is still working: it waits out *Acknowledgment Timeout*, thirty seconds by default and
 never less than ten. *Negative Acknowledgment Redelivery Delay* controls how soon the redelivery
-comes; it defaults to one second, and cannot be set longer than *Acknowledgment Timeout*. The
-reason for that rule: once a message is negatively acknowledged the client stops tracking it for
-the timeout, so the delay is the **only** thing that redelivers it — a delay longer than the timeout
-would make a write failure wait longer than a plain rollback did.
+comes; it defaults to five seconds. Keep it under *Acknowledgment Timeout*: once a message is
+negatively acknowledged the client stops tracking it for the timeout, so the delay is the **only**
+thing that redelivers it, and a longer delay makes a write failure wait longer than a plain rollback
+did. A longer delay is still accepted — it can be a deliberate backoff — but the processor logs a
+warning when it starts. Every redelivery also counts against *Max Redelivery Count*, so the delay
+sets how fast a message that keeps failing reaches the dead letter topic.
 
 > **Behaviour change since `2.11.0`:** in `2.11.0` the delay defaulted to Pulsar's own one minute,
 > so with *Acknowledgment Timeout* at its 30-second default a message the processor could not write
 > came back after **60 s — twice as long as before negative acknowledgement existed**, not sooner.
-> The default is now one second, so a flow that never set the property redelivers after a write
-> failure in about a second instead of a minute. A flow that had set the property to a value above
-> its *Acknowledgment Timeout* is now invalid and has to lower one or raise the other. To keep the
-> old gap on purpose, set the delay explicitly to a value no longer than the timeout.
+> The default is now five seconds, so a flow that never set the property redelivers after a write
+> failure in seconds instead of a minute. Two consequences. A flow with *Max Redelivery Count* set
+> now uses up its redeliveries **twelve times faster**: a transient write failure that used to be
+> absorbed by a minute per attempt can now send a perfectly good message to the dead letter topic
+> in seconds, so raise the count — or set the delay back up — to keep the retry window you had. And
+> a delay longer than *Acknowledgment Timeout* still validates, but the processor now logs a warning
+> when it starts; nothing that ran on `2.11.0` stops running.
 
 A message routed to `parse_failure` is **not** redelivered. It was delivered and handled — the
 flow has its bytes and can route them anywhere, including back to a Pulsar topic — so nacking it
--- a/docs/release-notes/2.11.0.1.md
+++ b/docs/release-notes/2.11.0.1.md
@@ -1,0 +1,38 @@
+# 2.11.0.1
+
+Connector revision for **NiFi 2.11.0** against the **Pulsar 4.2.4** client. Java 21. Unreleased;
+this file collects the notes for the next tag on the `2.11.0` line.
+
+## Behaviour changes
+
+**The default *Negative Acknowledgment Redelivery Delay* is five seconds, down from one minute
+(#218).** A negatively acknowledged message is redelivered by that delay and by nothing else: the
+Pulsar client drops it from the acknowledgment-timeout tracker the moment it is nacked. So with the
+one-minute default of `2.11.0` and *Acknowledgment Timeout* at its 30-second default, a message the
+processor could not write came back after sixty seconds — twice as long as before negative
+acknowledgement existed, when the timeout alone redelivered it in thirty. A flow that never set the
+property now redelivers after a write failure in about five seconds.
+
+Two things follow for flows that keep the default.
+
+- Redelivery is what advances *Max Redelivery Count*, so a flow with a dead letter policy uses up
+  its redeliveries about twelve times faster than on `2.11.0`. A transient write failure — a content
+  repository full for ten seconds — that used to be absorbed by a minute per attempt can now push
+  a perfectly good message to the dead letter topic within seconds. Raise *Max Redelivery Count*,
+  or set the delay explicitly, to keep the retry window the flow had.
+- The write-failure error is logged once per redelivery, so during an incident it repeats every
+  five seconds per message rather than every minute.
+
+**A delay longer than *Acknowledgment Timeout* is warned about, not rejected.** The processor logs
+a warning when it is scheduled with that combination, because the timeout no longer applies to a
+nacked message and the flow will wait longer after a write failure than a plain rollback did. It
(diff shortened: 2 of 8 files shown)


david-streamlio:
Reviewed at high effort. **The diagnosis is right and the core fix is correct** — I checked
`ConsumerImpl.negativeAcknowledge` on 4.2.4 and it does call `unAckedMessageTracker.remove(msgId)`,
so *Acknowledgment Timeout* genuinely stops applying to a nacked message and the `1 min` default did
invert what #216 set out to do. Thanks for catching this and for measuring it against both trees.

Four things before this merges.

### 1. The `1 sec` default speeds the dead-letter clock by 60× (medium)

`AbstractPulsarConsumerProcessor.java:216`. *Max Redelivery Count* is advanced **by redelivery**, and
redelivery is driven by this delay — this PR's own IT relies on exactly that relationship.

Scenario: a flow with `MAX_REDELIVER_COUNT=3` and a content repository that is full for ~10 seconds.
On `main` those messages get ~3 minutes of retry headroom and survive. After this change they are
redelivered three times within seconds and the client ships **perfectly good messages to the dead
letter topic** — the same transient write failure the nack path exists to serve.

That is a new regression traded for the old one. A default under the 10 s *Acknowledgment Timeout*
floor but well above 1 s (5 s, say) restores the promised behaviour without collapsing the DLQ
budget. Whatever the value, the README "Behaviour change" callout should say that *Max Redelivery
Count* now burns through far faster and suggest raising it.

Secondary effect of the same line: during a full-disk incident each undeliverable message logs
`"Unable to write the received messages to a FlowFile"` at ERROR **once per second per message**
instead of once per minute — adding write pressure to a disk that is already full.

### 2. The hard rejection breaks existing 2.11.0 flows on upgrade (medium)

`AbstractPulsarConsumerProcessor.java:463`. The new rule couples two knobs that were independent.

Scenario: an operator on 2.11.0 deliberately set the delay to `5 min` as a poison-message backoff,
leaving *Acknowledgment Timeout* at `30 sec`. After upgrading, `customValidate` marks the processor
invalid, it will not start, and the flow stalls until someone opens the config dialog. The same hits
anyone who explicitly pinned the documented `1 min` default.

It also removes long-backoff as a capability — and per finding 1, a longer delay is exactly the lever
someone would reach for to stop the DLQ burning too fast. Worth considering a validation *warning*
rather than an error. Either way this needs an upgrade note; `docs/release-notes/` has no file for
the next version yet, and `2.11.0.md:44` still documents the one-minute default.

### 3. The new IT can go green with no negative acknowledgement at all (medium)

`ConsumePulsarDeadLetterIT.java:169`. `failWriteOfNextMessage()` infers "received and refused" from
`!runner.getLogger().getErrorMessages().isEmpty()`. That list is **cumulative for the whole test and
never reset**, and the check is not specific to the write failure.

`ConsumePulsar.consume` catches `PulsarClientException`, logs `"Error communicating with Apache
Pulsar"` and returns without rethrowing (`ConsumePulsar.java:321-327`). So a transient client hiccup
during the preceding healthy `runner.run(...)` leaves an error in the list without failing the run.
The helper then returns on its first iteration — before the message reaches the receiver queue and
before any nack — `nackedAt` is stamped early, and the `await` loop's healthy pass consumes the
message directly. Green, with the nack path never exercised.

Fix: snapshot `getErrorMessages().size()` before the loop and require it to grow, or match on the
`"Unable to write the received messages"` text.

I'd flag that this is the same shape as the flaw this PR is fixing — the original IT pinned timings
that made its own assertion vacuous. Worth being strict here specifically.

### 4. The DLQ test now depends on a default it isn't testing (low)

`ConsumePulsarDeadLetterIT.java:68`. Removing `NEGATIVE_ACK_REDELIVERY_DELAY` from `init()` makes
*both* tests inherit the production default, but only the first is about the default.
`aRepeatedlyUndeliverableMessageEndsUpOnTheDeadLetterTopic` polls with `receive(2, SECONDS)` under a
60 s deadline, so any default above ~2 s — still legal under the new rule, whose ceiling is the 10 s
floor — makes it flake for reasons unrelated to what it tests. Set it explicitly there.

Also the two comment blocks at 68-71 have merged into something that no longer parses: "…the default
is the value under test. **The floor the validator allows.** The point of the redelivery test is…" —
the second sentence lost the `ACK_TIMEOUT` line it used to introduce.

jselvareef:
Thanks — all four taken, in `d1cad62`. PR body revised to match.

**1. Default is `5 sec`.** Agreed that `1 sec` traded one regression for another: a 10 s content-repository stall with `MAX_REDELIVER_COUNT=3` dead-letters good messages. `5 sec` keeps the invariant (under the 10 s floor, so every valid flow still redelivers a write failure sooner than the timeout alone would) and leaves the DLQ budget at 12× faster than `2.11.0` rather than 60×. The README callout, both `additionalDetails` and the property description now say that each redelivery advances *Max Redelivery Count* and suggest raising it; the release notes spell out the per-message log frequency as well.

**2. Warning, not rejection.** The validation rule is gone; a delay longer than *Acknowledgment Timeout* validates and starts, and `@OnScheduled` logs one warning naming both properties and values. `ValidationResult` has no warning level, so scheduling is the nearest hook that fires once per start rather than on every UI refresh. Two of the unit tests changed from `assertNotValid()` to "exactly one warning on `run(1, false, true)`", plus a new case asserting no warning for a within-timeout configuration. Upgrade note: new `docs/release-notes/2.11.0.1.md` (next connector revision per VERSIONING.md) with the behaviour change and an *Upgrading* section; I left `2.11.0.md`'s text as the record of that release and added a one-line pointer under the paragraph you cited.

**3. `failWriteOfNextMessage` is specific and monotonic.** It snapshots the count of errors whose message contains `"Unable to write the received messages"` before the loop and returns only when that count grows. Your point about the shape is fair — the fix was replacing one vacuous test with a subtler one, and the swallowed `"Error communicating with Apache Pulsar"` is exactly the error that would have triggered it.

**4. DLQ test sets `1 sec` explicitly** and the `init()` comment blocks are two paragraphs again, with the `ACK_TIMEOUT` sentence back where it belongs.

Verification on the new head: 356/356 unit; `ConsumePulsarDeadLetterIT` 2/2 against a real broker (32.6 s with the 5 s default). The rest of the IT suite was not re-run for this commit; the changes touch only the default value, `@OnScheduled`, and this IT.

PR #221, opened 2026-09-05
fix: hold one producer per topic under the exclusive access modes (#219)

### LINKED ISSUE #219: PublishPulsarRecord with Concurrent Tasks > 1 locks itself out under the exclusive Producer Access Modes; WaitForExclusive hangs the task
The three exclusive values of *Producer Access Mode* (#216) do not survive contact with `PublisherPool`. A single `PublishPulsarRecord` with **Concurrent Tasks = 2**, and nobody else writing the topic, locks itself out: with `Exclusive` half of the FlowFiles go to `failure`, with `ExclusiveWithFencing` the pool's producers fence each other, and with `WaitForExclusive` the second task blocks forever inside `onTrigger`.

### Where

`PublisherPool.obtainPublisher()` hands out an idle lease when there is one and otherwise creates a **new producer** for the topic. A returned lease goes back to the idle queue with its producer still open. `PublishPulsarRecord.onTrigger()` obtains a lease per trigger and returns it at the end, so two tasks that trigger at the same time each find the idle queue empty and each create a producer — two producers on one topic from one processor. That is exactly what the exclusive modes exist to refuse.

`PublishPulsar` is not affected: it shares one `currentLease` across tasks under `synchronized`, so it never holds two producers for the same topic. A single-task `PublishPulsarRecord` is fine too. The problem needs `PublishPulsarRecord` and Concurrent Tasks > 1, which is the normal way to run a publisher.

### Reproduction

`PublishPulsarRecord` against Testcontainers `apachepulsar/pulsar:4.2.4`, `runner.setThreadCount(2)`, two FlowFiles per round each carrying one record large enough (600 KB) that the 1 MB size filter gives each trigger exactly one FlowFile, so both tasks publish concurrently. Nothing else touches the topic.

| Mode | 10 FlowFiles, 5 rounds | What the log shows |
|---|---|---|
| `Exclusive` | 5 `success`, **5 `failure`** | `Unable to create producer` — `ProducerFencedException: Topic has an existing exclusive producer: standalone-0-5` (the pool's own producer) |
| `ExclusiveWithFencing` | 5 `success`, **5 `failure`** | `ProducerFencedException: Topic epoch has already moved. Current epoch: 1, Producer epoch: 0`, then `Producer was fenced` on every later send from the fenced producer, which the pool keeps handing out |
| `WaitForExclusive` | 1 `success`, run never finishes | second task parked in `ProducerBuilderImpl.create` ← `PublisherPool.createLease` ← `PublisherPool.obtainPublisher` ← `PublishPulsarRecord.onTrigger`; still there after 45 s, released only by closing the client |

The stuck thread for `WaitForExclusive`:

```
PROBE_STUCK thread=pool-3-thread-1 state=WAITING
    at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2072)
    at org.apache.pulsar.common.util.FutureUtil.getAndCleanupOnInterrupt(FutureUtil.java:414)
    at org.apache.pulsar.client.impl.ProducerBuilderImpl.create(ProducerBuilderImpl.java:87)
    at org.apache.nifi.processors.pulsar.utils.PublisherPool.createLease(PublisherPool.java:144)
    at org.apache.nifi.processors.pulsar.utils.PublisherPool.obtainPublisher(PublisherPool.java:111)
    at org.apache.nifi.processors.pulsar.pubsub.PublishPulsarRecord.onTrigger(PublishPulsarRecord.java:168)
```

`PublishPulsarProducerConfigurationIT` proves that an exclusive producer refuses a second *client*; it does not run a second task of the same processor, which is where this lives.

### Impact

- `Exclusive`: intermittent `failure` FlowFiles with a misleading cause — the "other producer" is the processor itself — for anyone who turns the property on with more than one task.
- `ExclusiveWithFencing`: the fenced producer stays in the idle queue and fails every send it is given from then on; each new producer fences the previous one, so the failures do not stop.
- `WaitForExclusive`: a NiFi thread pinned indefinitely inside `onTrigger`, with no timeout on `create()`. The processor cannot be stopped cleanly while that thread is parked.

### Options

- When the access mode is not `Shared`, have the pool keep **one producer per topic**: a second `obtainPublisher()` for a topic whose only lease is in use waits for it (bounded) instead of creating another. The exclusive modes then mean what they say for other flows and cannot be defeated from inside.
- Never block a NiFi thread on `create()` without a bound: `createAsync()` with a timeout, and route to `failure` when it fires — this matters for `WaitForExclusive` whatever else is done.
- At minimum, state in the property description, the *Producer behaviour* section and the README that the exclusive modes require Concurrent Tasks = 1 on `PublishPulsarRecord`, and reject them at validation on that processor if the framework's task count cannot be seen there.

Refs #186, #216.

jselvareef:
Closes #219.

`PublisherPool` created a producer for every lease that was in use at the same time. Under an exclusive *Producer Access Mode* that is the one thing the broker will not accept, so a `PublishPulsarRecord` with more than one concurrent task — and nobody else on the topic — collided with its own pool: `Exclusive` sent half of the FlowFiles to `failure` ("Topic has an existing exclusive producer" — its own), `ExclusiveWithFencing` had the pool's producers fence each other, and `WaitForExclusive` parked the second task inside `ProducerBuilderImpl.create()` for good.

> **Revised after review** (second commit): the wait for a held producer is bounded and its expiry sends the batch back to the queue rather than to `failure`; permits are keyed by the fully-qualified topic name; the permit release sits in a `finally`. Details in the review thread.

### What changed

- **`PublisherPool`** — when `accessMode` in the producer configuration is anything but `Shared`, the pool holds **one lease per topic**, the topic being `TopicName.get(name).toString()` so that `my-topic` and `persistent://public/default/my-topic` share it. A caller who finds the topic's lease in use waits (a fair per-topic `Semaphore`, `tryAcquire`) for at most `DEFAULT_EXCLUSIVE_LEASE_WAIT` = 5 s and is then handed the same producer; when the wait runs out or is interrupted it throws `PublisherUnavailableException` (new), which **`PublishPulsar` and `PublishPulsarRecord`** turn into "this FlowFile and the rest of the batch go back to the incoming queue, warn, `yield()`" — nothing was attempted for them, so `failure` is the wrong route. A creation that fails releases the topic so the next caller is not locked out, and the release on lease close is in a `finally`. A lease closed twice is now ignored instead of being queued twice — under exclusive access a double return would have handed the topic to two tasks at once, and under `Shared` it put the same producer in the idle queue twice. `Shared` is otherwise unchanged: concurrent tasks still get concurrent producers.
- **`PublisherPoolExclusiveAccessTest`** (new, mocked client, 9 cases) — for each of the three exclusive modes: the second task waits while the first holds the lease, receives that same lease once it is returned, and the builder sees exactly one `create()`. Plus: `Shared` still creates per concurrent lease, an absent `accessMode` is `Shared`, a double close releases the topic once, a failed creation does not leave the topic locked, the two spellings of one topic share its only producer (exclusive) and its idle lease (shared), the wait is bounded and giving up does not consume the permit, interruption is reported and the flag preserved.
- **`PublishPulsarHeldProducerTest`** (new) — both publishers, a pool whose producer is held: three FlowFiles enqueued, none in `success` or `failure`, all three back in the queue, `yield()` called, one warning, no error.
- **`PublishPulsarExclusiveAccessIT`** (new, real broker) — `PublishPulsarRecord` with two concurrent tasks, ten FlowFiles, all three modes: every FlowFile reaches `success`, none reaches `failure`, every record is on the topic, and the run finishes (the `WaitForExclusive` case is guarded by a timeout, which is the assertion there). Each FlowFile carries one 600 KB record so the 1 MB size-based filter hands a single FlowFile to each trigger and the two tasks genuinely publish at the same time.
- **README** — the *Producer Access Mode* paragraph says the processor keeps one producer per topic under the exclusive modes, with a *Behaviour change since `2.11.0`* callout; both publishers' `additionalDetails.html` say the same.

### Fails first

Unit tests on `main` (`00a6eb8`):

```
Tests run: 5, Failures: 2, Errors: 0, Skipped: 0 -- in PublisherPoolExclusiveAccessTest
  aSecondTaskWaitsForTheTopicsOnlyProducerInsteadOfCreatingAnother:
    Exclusive: the second task was handed a lease while the topic's producer was still in use
    - a second producer was created on an exclusive topic
  closingALeaseTwiceReleasesTheTopicOnce: expected:<1> but was:<2>
```

The IT on `main`, against `apachepulsar/pulsar:4.2.4`:

```
Tests run: 3, Failures: 2, Errors: 1, Skipped: 0, Time elapsed: 129.5 s -- in PublishPulsarExclusiveAccessIT
  twoTasksShareTheExclusiveProducer     -- expected: <0> but was: <5>   (FlowFiles in failure)
  twoTasksDoNotFenceEachOther           -- expected: <0> but was: <5>
  twoTasksDoNotWaitForEachOtherForever  -- TestTimedOutException: test timed out after 120000 milliseconds
```

With the fix:

```
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 17.17 s -- in PublishPulsarExclusiveAccessIT
```

Second commit, the new tests against the pool as it was after the first commit (raw-string keys, unbounded `acquire()`) and against routing to `failure`:

```
Tests run: 9, Failures: 2, Errors: 1 -- in PublisherPoolExclusiveAccessTest
  twoSpellingsOfOneTopicShareItsOnlyProducer: the fully-qualified spelling got its own producer while the
    short spelling's was in use
  anIdleProducerIsReusedUnderEitherSpelling: expected same:<...PooledPublisherLease@6438a7fe> was not:<...@4f5f6e45>
  aTaskDoesNotWaitForeverForAHeldProducer: TestTimedOut test timed out after 10000 milliseconds
Tests run: 2, Failures: 2 -- in PublishPulsarHeldProducerTest
  publishPulsarReturnsTheBatchToTheQueueAndYields: expected: <0> but was: <3>          (FlowFiles in failure)
  publishPulsarRecordReturnsTheBatchToTheQueueAndYields: expected: <0> but was: <3>
```

### Verification

- `mvn test` on the current head: **360/360** (349 + 9 + 2).
- Second commit: `PublishPulsar*IT` (8 classes, 25 tests) and `NiFiRoundTripIT` against a real broker: 26 pass; the 2 errors are the podman `execInContainer` broken pipe described below, in `PublishPulsarRoutingModeIT.roundRobinSpreadsUnkeyedMessagesOverEveryPartition` and `NiFiRoundTripIT.bytesSurviveTheRoundTrip`, thrown from `ExecCreateCmd` before any processor runs, reproducible here on every run today. The other tests in both classes pass.
- First commit: `mvn -B -ntp verify -Dgpg.skip=true -pl nifi-pulsar-processors -am` (the CI command) against a real broker via Testcontainers on rootless podman: 354 unit, **74 integration tests, 71 pass in the full run**. The 3 errors are `java.io.IOException: Broken pipe` from podman's exec endpoint inside `execInContainer` (`pulsar-admin` creating partitioned topics), thrown before any processor runs — an environment problem here, not the tests. Re-running those classes on their own, `ConsumePulsarRecordIT`, `NiFiRoundTripIT` and `PublishPulsarRoutingModeIT` all pass, so every integration test has passed against the branch at least once; CI's Docker should not see the exec issue at all.

### Not done

- The 5 s wait is a constant, not a property. It is the bound on a sibling's trigger, and the retry costs one yield; if a flow needs to tune it that can become a property later without changing the behaviour on expiry.
- The `finally` around the permit release has no test. `PublisherLease.close()` catches `PulsarClientException`, and the client's `flush()`/`close()` throw nothing else, so there is no path today that reaches the `finally` with an exception; the change is defensive.
- `WaitForExclusive` against an **external** holder still blocks inside `create()` until that holder disconnects — that is the mode's contract ("queues until it can take over") and is unchanged. What this PR removes is the case where the holder was the processor itself.
- `PublishPulsar` was not affected (it shares one `currentLease` across tasks) and is not changed beyond going through the same pool code.
- The access mode is read from the configuration map at pool construction; the processor puts the `ProducerAccessMode` enum there, and a string with the same name is accepted too, but I did not add a test for the string form.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,18 @@ Pulsar's broker-side deduplication requires.
 *Producer Access Mode* is how you stop two flows writing the same topic. `Shared`, the default,
 lets any number of producers write. `Exclusive` fails at producer creation if another producer
 already holds the topic; `WaitForExclusive` queues until it can take over; `ExclusiveWithFencing`
-evicts the incumbent and takes the topic.
+evicts the incumbent and takes the topic. Under any of the three the processor keeps **one
+producer per topic**, whatever its Concurrent Tasks: a task that needs a topic whose producer is
+busy waits for it instead of opening a second one, so the exclusivity is held against other flows
+and never turned against the processor itself.
+
+> **Behaviour change since `2.11.0`:** in `2.11.0` the publisher pool opened one producer per
+> concurrently held lease, so `PublishPulsarRecord` with more than one Concurrent Task collided
+> with its own producers under the exclusive modes: with `Exclusive` part of the FlowFiles went to
+> `failure` ("Topic has an existing exclusive producer" — its own), with `ExclusiveWithFencing`
+> the producers fenced each other, and with `WaitForExclusive` the second task blocked inside
+> `onTrigger` for good. Those flows now publish everything through the topic's single producer.
+> `Shared` is unchanged: concurrent tasks still get concurrent producers.
 
 *Batch Builder* decides how messages are grouped when *Batching Enabled* is on. `Default` fills a
 batch with whatever is pending, interleaving keys. **`Key based` is required for per-key ordering
--- a/README.md
+++ b/README.md
@@ -188,15 +188,22 @@ already holds the topic; `WaitForExclusive` queues until it can take over; `Excl
 evicts the incumbent and takes the topic. Under any of the three the processor keeps **one
 producer per topic**, whatever its Concurrent Tasks: a task that needs a topic whose producer is
 busy waits for it instead of opening a second one, so the exclusivity is held against other flows
-and never turned against the processor itself.
+and never turned against the processor itself. The topic is the topic as the broker sees it, so
+`my-topic` and `persistent://public/default/my-topic` share one producer. The wait is bounded — five
+seconds — because the task holding the producer may be inside a send that *Send Timeout* `0` lets
+run indefinitely; when it runs out, the FlowFiles of that trigger go **back to the incoming queue**,
+not to `failure`, the processor logs a warning and yields, and they are retried on a later trigger.
+Nothing was attempted for them, so nothing was refused.
 
 > **Behaviour change since `2.11.0`:** in `2.11.0` the publisher pool opened one producer per
 > concurrently held lease, so `PublishPulsarRecord` with more than one Concurrent Task collided
 > with its own producers under the exclusive modes: with `Exclusive` part of the FlowFiles went to
 > `failure` ("Topic has an existing exclusive producer" — its own), with `ExclusiveWithFencing`
 > the producers fenced each other, and with `WaitForExclusive` the second task blocked inside
-> `onTrigger` for good. Those flows now publish everything through the topic's single producer.
-> `Shared` is unchanged: concurrent tasks still get concurrent producers.
+> `onTrigger` for good. Those flows now publish everything through the topic's single producer,
+> and a task that cannot get it within five seconds returns its FlowFiles to the queue and yields
+> rather than failing them or waiting without limit. `Shared` is unchanged: concurrent tasks still
+> get concurrent producers.
 
 *Batch Builder* decides how messages are grouped when *Batching Enabled* is on. `Default` fills a
 batch with whatever is pending, interleaving keys. **`Key based` is required for per-key ordering
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarProducerProcessor.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarProducerProcessor.java
@@ -36,6 +36,7 @@ import org.apache.nifi.processor.*;
 import org.apache.nifi.processor.util.StandardValidators;
 import org.apache.nifi.processors.pulsar.utils.PropertyMappingUtils;
 import org.apache.nifi.processors.pulsar.utils.PublisherPool;
+import org.apache.nifi.processors.pulsar.utils.PublisherUnavailableException;
 import org.apache.nifi.pulsar.PulsarClientService;
 import org.apache.nifi.pulsar.cache.PulsarConsumerLRUCache;
 import org.apache.pulsar.client.api.CompressionType;
@@ -494,6 +495,25 @@ public abstract class AbstractPulsarProducerProcessor<T> extends AbstractProcess
 
     protected synchronized void setPublisherPool(PublisherPool pool) {
         this.publisherPool = pool;
+    }
+
+    /**
+     * The topic's only producer is with another task and did not come back within the pool's wait, so nothing was
+     * attempted for this FlowFile - or for the rest of the batch, which needs the same lease or one just as busy.
+     * They go back to the queue they came from, not to {@code failure}: a FlowFile routed to failure was tried and
+     * refused, and these were not. The yield keeps the next trigger from spinning on the same held producer.
+     */
+    protected void returnToQueueAndYield(final ProcessContext context, final ProcessSession session,
+                                         final FlowFile flowFile, final Iterator<FlowFile> rest,
+                                         final PublisherUnavailableException cause) {
+        int returned = 1;
+        session.transfer(flowFile);
+        while (rest.hasNext()) {
+            session.transfer(rest.next());
+            returned++;
+        }
+        getLogger().warn("{}; {} FlowFile(s) returned to the queue to be retried", cause.getMessage(), returned);
+        context.yield();
     }
 
     protected byte[] getDemarcatorBytes(ProcessContext context, final FlowFile flowFile) {
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/PublishPulsar.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/pubsub/PublishPulsar.java
@@ -37,6 +37,7 @@ import org.apache.nifi.processor.exception.ProcessException;
 import org.apache.nifi.processors.pulsar.AbstractPulsarProducerProcessor;
 import org.apache.nifi.processors.pulsar.utils.PublishPulsarUtils;
 import org.apache.nifi.processors.pulsar.utils.PublisherLease;
+import org.apache.nifi.processors.pulsar.utils.PublisherUnavailableException;
 
 @SeeAlso({ConsumePulsar.class, ConsumePulsarRecord.class, PublishPulsarRecord.class})
 @Tags({"Apache", "Pulsar", "Put", "Send", "Message", "PubSub"})
@@ -72,7 +73,13 @@ public class PublishPulsar extends AbstractPulsarProducerProcessor<byte[]> {
             final String topicName = context.getProperty(TOPIC).evaluateAttributeExpressions(flowFile).getValue();
             final boolean asyncFlag = (context.getProperty(ASYNC_ENABLED).isSet() && context.getProperty(ASYNC_ENABLED).asBoolean());
 
-            PublisherLease lease = obtainPublisherLease(topicName);
+            final PublisherLease lease;
+            try {
+                lease = obtainPublisherLease(topicName);
+            } catch (final PublisherUnavailableException e) {
+                returnToQueueAndYield(context, session, flowFile, itr, e);
+                return;
+            }
 
             if (lease == null) {
                 getLogger().error("Unable to publish to topic {}", new Object[] {topicName});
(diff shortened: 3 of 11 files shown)


david-streamlio:
Reviewed at high effort — one careful pass over `PublisherPool` plus the new tests, with the callers
(`AbstractPulsarProducerProcessor`, `PublishPulsar`, `PublishPulsarRecord`, `PublisherLease`) read for
context. `PublisherPool*Test` is 16/16 on the branch.

**The approach is sound.** The fair per-topic `Semaphore` plus the `leased` CAS guard does make the
pool hold one producer per topic under the three exclusive modes, and permit acquire/release is
balanced on every path I traced: idle reuse, `PulsarClientException`, the `IllegalStateException`
from a pool closed mid-`createLease`, and pool-close-then-lease-close. The double-close fix is a
genuine bug in its own right and helps `Shared` too. Docs are accurate.

Four findings, and the third means #219 is narrowed rather than closed.

### 1. Unbounded `permit.acquire()` inside `onTrigger`, with no release on close (high)

`PublisherPool.java:141`. `close()` never releases or interrupts waiters.

Scenario: one task holds the permit and its send is stuck — *Send Timeout* `0`, **which our own
README recommends for broker-side deduplication**, means a send can wait indefinitely. The sibling
tasks park inside `onTrigger` on flow-wide NiFi threads, holding uncommitted sessions, and stopping
the processor does not free them. The flow needs Terminate.

The PR lists the unbounded wait under "Not done", so I think we agree on the shape; I'd treat it as
blocking rather than follow-up, because the trigger condition is a configuration we actively
recommend. `tryAcquire` with a timeout, then `yield()` and roll back, resolves this and finding 4
together.

### 2. `release()` outside `finally` can permanently lose a topic's only permit (medium)

`PublisherPool.java:271`. `topicPermits.get(topicName).release()` sits after the close call rather
than in a `finally`. `PublisherLease.close()` only catches `PulsarClientException`, so any unchecked
exception out of `closeProducer()` skips the release. With one permit per topic, that topic is then
blocked for every later task for the life of the pool — a permanent stall from a transient fault.
One-line fix.

### 3. Permits are keyed by the raw topic string, so #219 still reproduces (high)

`PublisherPool.java:134`. Both the permit map and the idle queues key off the topic string exactly as
given. *Topic* supports Expression Language, so the same physical topic reached as `my-topic` and as
`persistent://public/default/my-topic` gets **two permits and two producers** — which is the original
bug, unchanged, just behind a narrower door.

That is not exotic: an EL expression deriving the topic from a FlowFile attribute is the normal way
to fan out, and Pulsar's own short form resolves to the fully-qualified name. Normalising through
`TopicName.get(topicName).toString()` before keying would close it, and the class already imports
Pulsar types so it costs nothing.

Worth a test with the two spellings of one topic, since that is the case the current IT cannot see.

### 4. Interruption while waiting sends an unattempted FlowFile to `failure` (medium)

`PublisherPool.java:143`. Interruption returns `null`, and `PublishPulsarRecord:170` /
`PublishPulsar:77` turn that into `session.transfer(flowFile, REL_FAILURE)`. A FlowFile that was never
attempted — the task was simply interrupted, typically on stop — is emitted down the failure path
rather than rolled back for retry. Bounding the wait per finding 1 and rolling back on both timeout
and interruption handles this.

---

Findings 1 and 3 are what I would want fixed before merge; 2 is a one-liner and 4 falls out of 1.
Thanks for tracking down the `WaitForExclusive` hang with a stack — that made the diagnosis obvious.

jselvareef:
Thanks — all four in `e920ce1`, PR body revised.

**1. Bounded wait.** `permit.tryAcquire(5 s)`; on expiry or interruption `obtainPublisher` throws a new `PublisherUnavailableException` rather than returning `null`, so the callers can tell "never attempted" from "topic blank / producer refused". Agreed it belonged in this PR given *Send Timeout* `0` is what the README recommends. The bound is a constant (`DEFAULT_EXCLUSIVE_LEASE_WAIT`) with a constructor overload for tests; I did not add a property, since the cost of expiry is one yield and a retry — easy to add later if a flow needs it.

**2. `finally`.** The release is in one now, after either return-to-idle or close-producer. No test for it: `PublisherLease.close()` catches `PulsarClientException` and the client's `flush()`/`close()` declare nothing else, so I could not find a path that reaches the `finally` with an exception; noted under "Not done" as defensive.

**3. Qualified topic keys.** Permits and idle queues are keyed by `TopicName.get(name).toString()`; a name the client cannot parse keeps its raw form and fails at producer creation as before. Two tests: `twoSpellingsOfOneTopicShareItsOnlyProducer` (exclusive — the fully-qualified spelling waits on the short one's lease and receives it) and `anIdleProducerIsReusedUnderEitherSpelling` (shared — the idle queue is shared too). Both fail against the first commit; output in the PR body.

**4. Interruption / expiry route.** Both `PublishPulsar` and `PublishPulsarRecord` catch the exception, `session.transfer(flowFile)` the current FlowFile and the rest of the batch back to the incoming queue, log one warning and `context.yield()`. `PublishPulsarHeldProducerTest` (new) runs each publisher with a pool whose producer is held: three FlowFiles in, none in `success` or `failure`, three back in the queue, yield called. Mutating the route to `failure` fails both cases (`expected: <0> but was: <3>`).

Verification: 360/360 unit; `PublishPulsar*IT` + `NiFiRoundTripIT` against a real broker, 26/28, the two errors being the podman `execInContainer` broken pipe in `PublishPulsarRoutingModeIT` and `NiFiRoundTripIT` (in `ExecCreateCmd`, before any processor runs — reproducible here today on every attempt, so I could not get those two individual tests green this time; the other tests in both classes pass, and CI's Docker did not see it on the first commit).

PR #222, opened 2026-09-05
feat: set an ordering key independently of the message key (#196)

### LINKED ISSUE #196: Allow an ordering key to be set independently of the message key
## Ask

Let the publish processors set a Pulsar **ordering key** separately from the message key.

## Where it stands today

Neither processor can. `PublisherLease` only ever sets the message key:

```java
TypedMessageBuilder tmb = producer.newMessage().properties(properties).value(value);
if (key != null) {
    tmb = tmb.key(key);
}
```

`orderingKey` appears nowhere in the codebase. `TypedMessageBuilder.orderingKey(byte[])` exists in the client and is unused, so the two are currently forced to be the same value.

## Why they are different

They drive different things:

| | Message key | Ordering key |
|---|---|---|
| Partition routing | yes | no |
| Topic compaction | yes | no |
| Key_Shared dispatch ordering | used when no ordering key is set | takes precedence |

Pulsar falls back to the message key when no ordering key is set, which is why nothing is broken today — but it means the two concerns cannot be separated when they differ.

They differ whenever the natural compaction key is not the natural ordering unit. Routing and compacting by tenant while ordering per session is the usual example: with only a message key you must choose which of the two to serve, and the other silently gets the wrong behaviour — messages compacted at the wrong granularity, or ordering guarantees that do not hold at the granularity the consumer assumes.

## Proposal

An **Ordering Key** property on `PublishPulsar` and an **Ordering Key Field** on `PublishPulsarRecord`, mirroring the existing *Message Key* / *Message Key Field* pair. When unset, nothing changes: Pulsar's own fallback to the message key applies, so no existing flow is affected.

An integration test would need a Key_Shared subscription to show it works — that the messages a consumer receives are ordered by the ordering key rather than the message key — which is not currently covered by any test.

## Scope note

This is a per-message concern, so it belongs on the processor rather than in the producer configuration list in #186. It is small and independent of #188.

## Related

- #186 — producer configuration properties; deliberately *not* where this belongs
- #190 — KeyValue SEPARATED topics, where the message key is owned by the schema; an ordering key would be unaffected by that and remains settable

### LINKED ISSUE #226: PublishPulsarRecord: Message Key Field on an Avro bytes field publishes an identity hash as the key
`PublishPulsarRecord` with *Message Key Field* naming an Avro `bytes` field publishes **`[Ljava.lang.Object;@5cf57368`** as the message key — the identity hash of the array object, different for every record. Nothing fails: the messages are keyed, routed and compacted by a value that is not the field's content and never repeats, so per-key ordering and compaction by that field are silently gone.

### Where

`PublisherLease.getMessageKey(FlowFile, RecordSetWriterFactory, Object)` converts the field value:

```java
} else if (keyValue instanceof byte[]) {
    messageKey = (byte[]) keyValue;
} else if (keyValue instanceof Byte[]) {
    // This case exists because in our Record API we currently don't have a BYTES type, we use an Array of type
    // Byte, which creates a Byte[] instead of a byte[]. ...
    ...
} else {
    final String keyString = keyValue.toString();
```

The `Byte[]` branch was written for the boxed shape the author expected. NiFi's `AvroTypeUtil` does not produce it: `convertByteArray(byte[])` returns **`Object[]`** — `anewarray java/lang/Object`, elements boxed as `Byte` — and that is what an `AvroReader` hands over for an Avro `bytes` field (checked on `nifi-avro-record-utils` 2.11.0 and 2.9.0). `Object[]` is not `instanceof Byte[]`, so the value falls through to `toString()`, and `Object[].toString()` is the identity hash. The branch reads correctly and, as far as I can tell, has never fired for Avro input.

### Reproduction

Unit level, no broker: `PublishPulsarRecord`, a real `AvroReader` (embedded schema), a record schema with `"session": {"type": "bytes"}`, *Message Key Field* = `session`, three records with sessions `s-1`, `s-1`, `s-2`. Expected `key("s-1")` twice and `key("s-2")` once; Mockito's report of the first send instead:

```
Argument(s) are different! Wanted:
typedMessageBuilder.key("s-1");
Actual invocations have different arguments:
typedMessageBuilder.key("[Ljava.lang.Object;@5cf57368");
```

Each record gets its own array object and therefore its own hash, so no two of the three keys are equal — the two `s-1` records do not share a key.

`MockRecordParser` cannot show this — it yields every field as a `String` — which is presumably why no existing test caught it.

### Scope

- Only *Message Key Field*, only when the field is an Avro `bytes` (any reader that goes through `AvroTypeUtil`, so `AvroReader` and a schema-registry-backed Avro read). `string` and numeric key fields are unaffected. `PublishPulsar`'s *Message Key* property is a string and unaffected.
- The published messages are valid and readable; the damage is to routing (a partitioned topic spreads records that share a real key over partitions at random) and to compaction (nothing is ever superseded, since no two messages share a key).

### Fix

Accept the shape actually produced: an `Object[]` whose every element is a `Byte`, alongside `Byte[]`, unboxed to the content. #222 does this as part of routing the ordering key field through the same conversion — the two properties then agree on every value — and adds the `AvroReader`-based test above. Filed separately, as requested there, so the fix is findable by someone who searches for the hash string.

Refs #222.

jselvareef:
Closes #196. Closes #226.

The publishers could only set the message key, so partition routing, compaction and `Key_Shared` dispatch were forced onto one value. Pulsar keeps them apart: the message key routes and compacts, the ordering key decides `Key_Shared` dispatch and takes precedence there when set. This adds the pair the issue proposes, and nothing else.

> **Revised after review** (second commit): the ordering key field is converted through the same helper as the message key field (so an Avro `bytes` field is its content, not an identity hash), blank means no key on both processors, *Ordering Key* is offered by `PublishPulsar` only, and the IT's control run is deterministic. **Third commit:** the *Message Key Field* fix that fell out of the shared helper is filed as #226, closed here, and carried as its own entry in the release notes with a README behaviour-change callout. **Fourth commit (second review):** a binary key field is sent as key bytes rather than decoded text, an empty array is not a byte string, a key field the schema lacks is warned about, and the IT's `Key_Shared` consumers own explicit sticky hash ranges. Details in the review thread.

### What changed

- **`PublishPulsar`** — *Ordering Key*, Expression Language against FlowFile attributes, mirroring *Message Key*. Blank (or an expression that resolves to nothing) means no ordering key is set, so Pulsar falls back to the message key exactly as before. No `msg.orderingKey` attribute fallback — an expression covers that case without inventing a second convention. The property lives on this processor, not the base class: `PublishPulsarRecord` never reads it, and would otherwise show it.
- **`PublishPulsarRecord`** — *Ordering Key Field*, mirroring *Message Key Field*. A record whose field is null or blank is sent without an ordering key; other records of the same FlowFile still carry theirs. The field yields the same bytes as it would under *Message Key Field* — `byte[]` as is, an array of boxed bytes unboxed (an empty array only if it is a `Byte[]`), a nested `Record` serialised with the FlowFile's writer, anything else as UTF-8 text. A **binary** message key field is sent through `keyBytes()` (base64 on the wire, flagged), never decoded into a `String`; a text one is read as UTF-8. A field name the records' schema does not have is warned about once per FlowFile, for both properties.
- **`PublisherLease`** — the key travels through every send path, including the KeyValue `SEPARATED` one, where the schema owns the message key but the ordering key is unaffected (as #196 notes). The existing `send`/`sendAsync`/`…WithKeyBytes` methods remain the extension point: the new ordering-key variants hand every message *without* an ordering key to them, so a subclass overriding the originals — the test suite does, to control the futures — keeps seeing those messages. The public `publish` overloads are kept and delegate.
- **Docs** — README's *Publisher message metadata* table gains the row and a paragraph on which key does what, with the reminder that *Batch Builder* = `Key based` is what keeps a batch from spanning keys; both publishers' `additionalDetails.html` say the same. A *Behaviour change since `2.11.0`* callout covers #226: a flow keying by an Avro `bytes` field starts routing and compacting by the real key, sent as a binary key. New `docs/release-notes/2.11.0.2.md` (the next connector revision under VERSIONING.md, `2.11.0.1` having shipped) lists the ordering key under *New* and #226 plus the missing-field warning under *Fixes*.

### Tests

**`PublishPulsarOrderingKeyTest`** (mocked client, 16 cases): the key set from the property and from an attribute expression, applied to every demarcated message, taken from the record field per record, skipped for a null/empty field, and — the fallback contract — never set when the property is blank or the expression resolves to nothing. Second commit: an Avro `bytes` field read by a real `AvroReader` yields the same ordering key for two records with the same content and the same bytes as the message key from the same field; a blank field sets no ordering key; *Ordering Key* is offered by `PublishPulsar` and not by `PublishPulsarRecord`. Fourth commit: two byte strings that decode to the same text (`FF FE 01` / `FE FF 01`) are sent as two different `keyBytes` and `key(String)` is never called; an empty `array<string>` field is not sent as key `""`; a misspelt *Ordering Key Field* or *Message Key Field* produces exactly one warning naming the property, the field and the schema's fields, and sets no key; a present field warns nothing.

**`PublishPulsarOrderingKeyIT`** (Testcontainers, `apachepulsar/pulsar:4.2.4`, 5 cases): the key arrives on the message alongside the message key; nothing is set when the property is unset; `PublishPulsarRecord` takes it from a field; and the one only a broker can answer — **on a two-consumer `Key_Shared` subscription, 40 messages with distinct message keys and one ordering key all land on a single consumer, in publish order**, while the control run with the same distinct message keys and no ordering key spreads them over both consumers. The control is what stops the assertion from also holding on a broker that ignored the ordering key. The two consumers own explicit sticky hash ranges (`KeySharedPolicy.stickyHashRange()`, `0–32767` and `32768–65535`), so which consumer a key lands on is `hash(key) % 65536` against those ranges on any broker rather than a property of the broker's AUTO_SPLIT ring; both runs assert that all 40 arrived before judging dispatch. Fourth commit adds the binary-key case: an Avro `bytes` key field arrives `hasBase64EncodedKey()` with `getKeyBytes()` equal to the field, and two byte strings that decode to the same text are two keys.

### Fails first

A new property has no "before" on `main` beyond a compile error, so the meaningful red is the mutation: with the ordering key dropped before `send` (the `orderingKey(...)` call removed) the 5 unit cases that assert the key fail and the 3 fallback cases keep passing —

```
Tests run: 8, Failures: 5, Errors: 0, Skipped: 0 -- in PublishPulsarOrderingKeyTest
  publishPulsarPutsTheOrderingKeyOnTheMessage:105->orderingKeysSent:88
  publishPulsarEvaluatesTheOrderingKeyAgainstFlowFileAttributes:117->orderingKeysSent:88
  publishPulsarAppliesTheOrderingKeyToEveryDemarcatedMessage:131->orderingKeysSent:88
  publishPulsarRecordTakesTheOrderingKeyFromTheNamedField:174->orderingKeysSent:88
  publishPulsarRecordSkipsTheOrderingKeyForARecordWithoutTheField:187->orderingKeysSent:88
```

— and the same mutation would fail the `Key_Shared` IT, since the distinct message keys then spread the messages.

Second commit, the Avro `bytes` case against the first commit's conversion (`toString()` on anything that is not a `byte[]`):

```
publishPulsarRecordUnboxesAnAvroBytesFieldLikeTheMessageKeyDoes
  typedMessageBuilder.orderingKey([... "[Ljava.lang.Object;@5cf57368" as bytes ...])
  typedMessageBuilder.key("[Ljava.lang.Object;@5cf57368")
```

Both keys were the identity hash — which is also why the fix is in the shared helper, see the thread.

Fourth commit, the five new unit cases against the third commit's `PublisherLease`:

```
Tests run: 16, Failures: 5 -- in PublishPulsarOrderingKeyTest
  aBinaryMessageKeyFieldIsSentAsKeyBytesSoDifferentValuesStayDifferentKeys:
    Wanted but not invoked: typedMessageBuilder.keyBytes(<Capturing argument: byte[]>)
  publishPulsarRecordUnboxesAnAvroBytesFieldUnderBothProperties:  (same: key(String) was used, not keyBytes)
  anEmptyArrayFieldIsNotABinaryKey:  NeverWantedButInvoked: typedMessageBuilder.key("")
  aMisspelledOrderingKeyFieldIsWarnedAboutOncePerFlowFile:  expected:<1> but was:<0>
  aMisspelledMessageKeyFieldIsWarnedAboutOncePerFlowFile:   expected:<1> but was:<0>
```

### Verification

- `mvn test` on the current head: **383/383** (on top of `main` after `2.11.0.1`: 367 + 16).
- Fourth commit: `PublishPulsarOrderingKeyIT` 5/5, `PublishPulsarMessageKeyIT` 4/4, `PublishPulsarRecordSchemaIT` 3/3, `KeyValueSchemaRoundTripIT` 7/7 against a real broker.
- Second commit: `PublishPulsarOrderingKeyIT` (three consecutive runs, 4/4 each), `PublishPulsarMessageKeyIT` 4/4 and `PublishPulsarRecordSchemaIT` 3/3 against a real broker.
- First commit:
- `mvn -B -ntp verify -Dgpg.skip=true -pl nifi-pulsar-processors -am` (the CI command) against a real broker via Testcontainers on rootless podman: 357 unit, **75 integration tests, 73 pass in the full run**. The 2 errors are `java.io.IOException: Broken pipe` from podman's exec endpoint inside `execInContainer` (`pulsar-admin` creating partitioned topics), thrown before any processor runs — the same environment problem noted on #220 and #221, not the tests. Re-run on their own, `NiFiRoundTripIT` and `PublishPulsarRoutingModeIT` both pass, so every integration test has passed against the branch at least once.

### Not done

- No behaviour-change callout for the ordering key itself: with the properties unset the message is byte-for-byte what it was. The one existing case the shared helper changes — *Message Key Field* naming an Avro `bytes` field — is #226, kept in this PR at the reviewer's request and documented as a fix in its own right.
- Release notes: `2.11.0.1` has shipped, so this PR's *New* (#196) and *Fixes* (#226) entries live in a new `docs/release-notes/2.11.0.2.md`; the released `2.11.0.1.md` is untouched. #228 creates the same file for its own entry; whichever lands second I rebase and fold together. Rebased onto `main` after `2.11.0.1` with no code conflicts.
- The nested-`Record` case of the ordering key goes through the same serialisation as the message key's, but has no test of its own.
- Text keys are now read as UTF-8 rather than the platform default charset. On a node whose `file.encoding` is not UTF-8 a non-ASCII text key changes to the correct, node-independent bytes; noted in the release notes.
- The missing-field warning checks the record set's schema. A reader that keeps unknown fields (`dropUnknownFields` off) can still supply a value for a name the schema lacks, in which case the warning is a false alarm; the message says "not a field of the records' schema", which stays true.
- The `Key_Shared` IT runs with batching off so each message is dispatched on its own key; the batched case is *Batch Builder*'s concern (#216) and is documented rather than tested here.
- `PublishPulsarLeaseReuseTest` verifies the processor's call into the lease and was updated to the new arity; that is the only existing test touched.
- The KeyValue `SEPARATED` path is covered by the unit-level plumbing, not by an IT with a KeyValue topic.

### DIFF
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/utils/PublisherLease.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/utils/PublisherLease.java
@@ -189,6 +189,16 @@ public class PublisherLease implements Closeable {
 
     public void publish(final FlowFile flowFile, final InputStream flowFileContent, final String messageKey,
                         Map<String, String> messageProperties, final byte[] demarcatorBytes, boolean async) throws IOException {
+        publish(flowFile, flowFileContent, messageKey, null, messageProperties, demarcatorBytes, async);
+    }
+
+    /**
+     * @param orderingKey Pulsar's ordering key for every message of the FlowFile, or {@code null} to set none, in
+     *                    which case the broker falls back to the message key
+     */
+    public void publish(final FlowFile flowFile, final InputStream flowFileContent, final String messageKey,
+                        final byte[] orderingKey, Map<String, String> messageProperties, final byte[] demarcatorBytes,
+                        boolean async) throws IOException {
 
         byte[] messageContent;
         List<CompletableFuture<MessageId>> futureList = new ArrayList<>();
@@ -197,16 +207,16 @@ public class PublisherLease implements Closeable {
             messageContent = new byte[(int) flowFile.getSize()];
             StreamUtils.fillBuffer(flowFileContent, messageContent);
             futureList.add(async ?
-                    sendAsync(producer, messageKey, messageProperties, messageContent) :
-                    send(producer, messageKey, messageProperties, messageContent));
+                    sendAsync(producer, messageKey, orderingKey, messageProperties, messageContent) :
+                    send(producer, messageKey, orderingKey, messageProperties, messageContent));
 
         } else {
             try (final StreamDemarcator demarcator = new StreamDemarcator(flowFileContent, demarcatorBytes, Integer.MAX_VALUE)) {
 
                 while ((messageContent = demarcator.nextToken()) != null) {
                     futureList.add(async ?
-                            sendAsync(producer, messageKey, messageProperties, messageContent) :
-                            send(producer, messageKey, messageProperties, messageContent));
+                            sendAsync(producer, messageKey, orderingKey, messageProperties, messageContent) :
+                            send(producer, messageKey, orderingKey, messageProperties, messageContent));
 
                     if (futureList.size() > 99) {
                         producer.flush();
@@ -248,6 +258,18 @@ public class PublisherLease implements Closeable {
                         final RecordSchema schema, final String messageKeyField, Map<String, String> messageProperties,
                         boolean async, boolean useTopicSchema, final String keyValueKeyField,
                         final String keyValueValueField) throws IOException {
+        publish(flowFile, recordSet, writerFactory, schema, messageKeyField, null, messageProperties, async, useTopicSchema,
+                keyValueKeyField, keyValueValueField);
+    }
+
+    /**
+     * @param orderingKeyField the record field whose value becomes the message's ordering key, or {@code null} to set
+     *                         none; a record whose field is null or empty gets no ordering key either
+     */
+    public void publish(final FlowFile flowFile, final RecordSet recordSet, final RecordSetWriterFactory writerFactory,
+                        final RecordSchema schema, final String messageKeyField, final String orderingKeyField,
+                        Map<String, String> messageProperties, boolean async, boolean useTopicSchema,
+                        final String keyValueKeyField, final String keyValueValueField) throws IOException {
 
         final TopicSchema resolvedSchema = useTopicSchema ? getTopicSchema() : null;
 
@@ -288,6 +310,8 @@ public class PublisherLease implements Closeable {
 
                 final byte[] messageContent;
                 final String messageKey;
+                final byte[] orderingKey = orderingKeyField == null || orderingKeyField.isEmpty()
+                        ? null : getOrderingKey(record.getValue(orderingKeyField));
 
                 if (keyValueSchema != null) {
                     final KeyValueTopicSchema.EncodedKeyValue encoded =
@@ -310,8 +334,8 @@ public class PublisherLease implements Closeable {
                         }
 
                         futureList.add(async
-                                ? sendAsyncWithKeyBytes(producer, encoded.getMessageKey(), messageProperties, messageContent)
-                                : sendWithKeyBytes(producer, encoded.getMessageKey(), messageProperties, messageContent));
+                                ? sendAsyncWithKeyBytes(producer, encoded.getMessageKey(), orderingKey, messageProperties, messageContent)
+                                : sendWithKeyBytes(producer, encoded.getMessageKey(), orderingKey, messageProperties, messageContent));
 
                         if (futureList.size() > 100) {
                             producer.flush();
@@ -340,8 +364,8 @@ public class PublisherLease implements Closeable {
                 messageKey = getMessageKey(flowFile, writerFactory, record.getValue(messageKeyField));
 
                 futureList.add(async ?
-                        sendAsync(producer, messageKey, messageProperties, messageContent) :
-                        send(producer, messageKey, messageProperties, messageContent));
+                        sendAsync(producer, messageKey, orderingKey, messageProperties, messageContent) :
+                        send(producer, messageKey, orderingKey, messageProperties, messageContent));
 
                 if (futureList.size() > 100) {
                     producer.flush();
@@ -580,11 +604,38 @@ public class PublisherLease implements Closeable {
         return producer.newMessage().properties(properties).keyBytes(keyBytes).value(value).sendAsync();
     }
 
+    /**
+     * The variants taking an ordering key hand a message without one to the original methods, so a subclass that
+     * overrides those (tests do, to control the futures) keeps seeing every message that has no ordering key.
+     */
+    protected CompletableFuture<MessageId> sendAsyncWithKeyBytes(Producer producer, byte[] keyBytes, byte[] orderingKey,
+                                                                 Map<String, String> properties, byte[] value) {
+        if (orderingKey == null) {
+            return sendAsyncWithKeyBytes(producer, keyBytes, properties, value);
+        }
+        return producer.newMessage().properties(properties).keyBytes(keyBytes).orderingKey(orderingKey).value(value).sendAsync();
+    }
+
     protected CompletableFuture<MessageId> sendWithKeyBytes(Producer producer, byte[] keyBytes, Map<String, String> properties, byte[] value)
             throws PulsarClientException {
         try {
             return CompletableFuture.completedFuture(
                     producer.newMessage().properties(properties).keyBytes(keyBytes).value(value).send());
+        } catch (final PulsarClientException e) {
+            final CompletableFuture<MessageId> failed = new CompletableFuture<>();
+            failed.completeExceptionally(e);
+            return failed;
+        }
+    }
+
+    protected CompletableFuture<MessageId> sendWithKeyBytes(Producer producer, byte[] keyBytes, byte[] orderingKey,
+                                                            Map<String, String> properties, byte[] value) throws PulsarClientException {
+        if (orderingKey == null) {
+            return sendWithKeyBytes(producer, keyBytes, properties, value);
+        }
+        try {
+            return CompletableFuture.completedFuture(
+                    producer.newMessage().properties(properties).keyBytes(keyBytes).orderingKey(orderingKey).value(value).send());
         } catch (final PulsarClientException e) {
             final CompletableFuture<MessageId> failed = new CompletableFuture<>();
             failed.completeExceptionally(e);
@@ -602,6 +653,37 @@ public class PublisherLease implements Closeable {
     }
 
     /**
+     * Sets the ordering key when there is one. Left unset, the broker falls back to the message key, which is what
+     * every message carried before the ordering key could be given separately (#196).
+     */
+    protected CompletableFuture<MessageId> sendAsync(Producer producer, String key, byte[] orderingKey, Map<String, String> properties, byte[] value) {
+        if (orderingKey == null) {
+            return sendAsync(producer, key, properties, value);
+        }
+        TypedMessageBuilder tmb = producer.newMessage().properties(properties).orderingKey(orderingKey).value(value);
+
+        if (key != null) {
+            tmb = tmb.key(key);
+        }
+        return tmb.sendAsync();
+    }
+
+    /**
+     * The ordering key a record field yields: bytes as they are, anything else as its UTF-8 string form, and no
+     * key at all for a null or empty value.
+     */
+    private static byte[] getOrderingKey(final Object fieldValue) {
+        if (fieldValue == null) {
+            return null;
+        }
+        if (fieldValue instanceof byte[]) {
+            return ((byte[]) fieldValue).length == 0 ? null : (byte[]) fieldValue;
+        }
+        final String asString = fieldValue.toString();
+        return asString.isEmpty() ? null : asString.getBytes(StandardCharsets.UTF_8);
+    }
+
+    /**
      * Sends one message and waits for it, on the calling thread.
      * <p>
      * This used to wrap the blocking send in {@code CompletableFuture.supplyAsync()}, handing every message
@@ -616,6 +698,25 @@ public class PublisherLease implements Closeable {
      */
     protected CompletableFuture<MessageId> send(Producer producer, String key, Map<String, String> properties, byte[] value) {
         TypedMessageBuilder tmb = producer.newMessage().properties(properties).value(value);
+
+        if (key != null) {
+            tmb = tmb.key(key);
+        }
+
+        try {
+            return CompletableFuture.completedFuture(tmb.send());
+        } catch (final PulsarClientException e) {
+            final CompletableFuture<MessageId> failed = new CompletableFuture<>();
+            failed.completeExceptionally(e);
+            return failed;
+        }
+    }
+
+    protected CompletableFuture<MessageId> send(Producer producer, String key, byte[] orderingKey, Map<String, String> properties, byte[] value) {
+        if (orderingKey == null) {
+            return send(producer, key, properties, value);
+        }
+        TypedMessageBuilder tmb = producer.newMessage().properties(properties).orderingKey(orderingKey).value(value);
 
         if (key != null) {
             tmb = tmb.key(key);
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/utils/PublisherLease.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/utils/PublisherLease.java
@@ -302,6 +302,12 @@ public class PublisherLease implements Closeable {
 
         final ByteArrayOutputStream baos = new ByteArrayOutputStream(1024);
 
+        // Once per FlowFile: a key field the records do not have yields null for every record, which is
+        // indistinguishable from a field that is present and null, so a misspelt name fails in silence. For
+        // the ordering key the only observable effect is dispatch on a multi-consumer subscription.
+        warnIfNotAField(recordSet.getSchema(), messageKeyField, "Message Key Field");
+        warnIfNotAField(recordSet.getSchema(), orderingKeyField, "Ordering Key Field");
+
         Record record;
         List<CompletableFuture<MessageId>> futureList = new ArrayList<>();
 
@@ -310,7 +316,6 @@ public class PublisherLease implements Closeable {
                 baos.reset();
 
                 final byte[] messageContent;
-                final String messageKey;
                 final byte[] orderingKey = orderingKeyField == null || orderingKeyField.isEmpty()
                         ? null : getOrderingKey(flowFile, writerFactory, record.getValue(orderingKeyField));
 
@@ -362,11 +367,26 @@ public class PublisherLease implements Closeable {
 
                     messageContent = baos.toByteArray();
                 }
-                messageKey = getMessageKey(flowFile, writerFactory, record.getValue(messageKeyField));
+                final Object keyValue = messageKeyField == null || messageKeyField.isEmpty() ? null : record.getValue(messageKeyField);
+                final byte[] messageKey = keyBytes(flowFile, writerFactory, keyValue);
 
-                futureList.add(async ?
-                        sendAsync(producer, messageKey, orderingKey, messageProperties, messageContent) :
-                        send(producer, messageKey, orderingKey, messageProperties, messageContent));
+                if (messageKey != null && isBinary(keyValue)) {
+                    // A byte field is a binary key and travels as one: keyBytes() carries the bytes as they are
+                    // (base64 on the wire, flagged as such), so two different values are always two different
+                    // keys. Decoding them into a String would go through a charset, and a charset maps every
+                    // invalid sequence to the same replacement character - two different UUIDs could become one
+                    // key, route to one partition and supersede each other on a compacted topic.
+                    futureList.add(async
+                            ? sendAsyncWithKeyBytes(producer, messageKey, orderingKey, messageProperties, messageContent)
+                            : sendWithKeyBytes(producer, messageKey, orderingKey, messageProperties, messageContent));
+                } else {
+                    // Text, a number, a nested record the writer serialised: a text key, always read as UTF-8 so
+                    // that every node of a cluster derives the same key from the same record.
+                    final String textKey = messageKey == null ? null : new String(messageKey, StandardCharsets.UTF_8);
+                    futureList.add(async
+                            ? sendAsync(producer, textKey, orderingKey, messageProperties, messageContent)
+                            : send(producer, textKey, orderingKey, messageProperties, messageContent));
+                }
 
                 if (futureList.size() > 100) {
                     producer.flush();
@@ -732,16 +752,12 @@ public class PublisherLease implements Closeable {
         }
     }
 
-    private String getMessageKey(final FlowFile flowFile, final RecordSetWriterFactory writerFactory,
-                                 final Object keyValue) throws IOException, SchemaNotFoundException {
-        final byte[] messageKey = keyBytes(flowFile, writerFactory, keyValue);
-        return (messageKey == null) ? null : new String(messageKey);
-    }
-
     /**
      * The bytes a record field contributes as a key. Shared by the message key and the ordering key, so a field
-     * of any type means the same thing under either property: bytes as they are, an array of boxed bytes unboxed,
-     * a nested record serialised with the FlowFile's writer, anything else as its UTF-8 string.
+     * of any type yields the same bytes under either property: bytes as they are, an array of boxed bytes unboxed,
+     * a nested record serialised with the FlowFile's writer, anything else as its UTF-8 string. What the caller
+     * does with them differs by necessity - the ordering key is always bytes, the message key is bytes for a
+     * binary field ({@link #isBinary}) and UTF-8 text otherwise - but the value is the same.
      */
     private byte[] keyBytes(final FlowFile flowFile, final RecordSetWriterFactory writerFactory,
                             final Object keyValue) throws IOException, SchemaNotFoundException {
@@ -774,10 +790,30 @@ public class PublisherLease implements Closeable {
         }
     }
 
-    /** True for an array whose every element is a {@link Byte} - a byte string in the Record API's clothing. */
+    private void warnIfNotAField(final RecordSchema recordSchema, final String fieldName, final String property) {
+        if (fieldName == null || fieldName.isEmpty() || recordSchema == null || recordSchema.getField(fieldName).isPresent()) {
+            return;
+        }
+        logger.warn("{} names '{}', which is not a field of the records' schema (fields: {}); every record of this "
+                + "FlowFile is sent as if the field were null", property, fieldName, recordSchema.getFieldNames());
+    }
+
+    /** Whether a key field's value is binary - a {@code byte[]}, or the boxed form the Record API uses for one. */
+    static boolean isBinary(final Object value) {
+        return value instanceof byte[] || (value instanceof Object[] && isBoxedBytes((Object[]) value));
+    }
+
+    /**
+     * True for an array whose every element is a {@link Byte} - a byte string in the Record API's clothing. An
+     * empty array is bytes only when it is declared as such: an empty {@code array<string>} field is also an
+     * empty {@code Object[]}, and must not turn into an empty key.
+     */
     private static boolean isBoxedBytes(final Object[] array) {
         if (array instanceof Byte[]) {
             return true;
+        }
+        if (array.length == 0) {
+            return false;
         }
         for (final Object element : array) {
             if (!(element instanceof Byte)) {
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/utils/PublisherLease.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/utils/PublisherLease.java
@@ -19,6 +19,7 @@ package org.apache.nifi.processors.pulsar.utils;
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
 import org.apache.commons.compress.utils.IOUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.nifi.avro.AvroTypeUtil;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericDatumWriter;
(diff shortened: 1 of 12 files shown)


david-streamlio:
Reviewed at high effort. Checked the PR head out in a worktree: compiles, 357 unit tests pass
(`PublishPulsarOrderingKeyTest` 8/8, `PublishPulsarLeaseReuseTest` 12/12).

**The plumbing is sound.** The new `send*` overloads delegate to the originals when
`orderingKey == null`, so a flow with the properties unset is byte-for-byte unchanged — which is the
property that matters most here. I also confirmed Pulsar's `KeyBasedBatchMessageContainer` hashes the
ordering key when one is present, so the "pair this with *Batch Builder* = `Key based`" guidance in
the docs is correct rather than folklore.

One finding I would fix before merge, and three smaller ones.

### 1. `getOrderingKey` doesn't handle `Byte[]`, producing a per-message garbage key (high)

`PublisherLease.java:675`. Its sibling `getMessageKey` (same file, line 741) has an explicit `Byte[]`
branch with the comment *"in our Record API we currently don't have a BYTES type, we use an Array of
type Byte, which creates a `Byte[]` instead of a `byte[]`"*. `getOrderingKey` checks only
`instanceof byte[]` and otherwise falls through to `fieldValue.toString()`.

Scenario: an Avro-read record with a `bytes` field (NiFi's `AvroTypeUtil` converts `ByteBuffer` →
`Byte[]`) named by *Ordering Key Field* yields the ordering key `"[Ljava.lang.Byte;@4f8e5cde"` — a
**different identity-hash string for every message**. Key_Shared dispatch then scatters those
messages across consumers instead of pinning them, which is the exact opposite of what the feature
exists to do. The same field used as *Message Key Field* works correctly, so the two properties
disagree on identical input.

The `Record` case has the same gap: `getMessageKey` serialises it via the record writer,
`getOrderingKey` emits `MapRecord[{...}]`.

Mirroring the existing `Byte[]` branch is the fix; worth a test with an Avro `bytes` field, since
that is what makes the divergence visible.

### 2. `PublishPulsarRecord` exposes an *Ordering Key* it never reads (medium)

`PublishPulsarRecord.java:155`. `properties.addAll(AbstractPulsarProducerProcessor.PROPERTIES)` pulls
in the new `ORDERING_KEY` (`AbstractPulsarProducerProcessor.java:370`), so the UI shows both *Ordering
Key* and *Ordering Key Field* on this processor — and *Ordering Key*'s description promises it "sets
Pulsar's ordering key on every message". `onTrigger` reads only `ORDERING_KEY_FIELD` (line 197);
`getOrderingKey(ProcessContext, FlowFile)` is never called from here.

A user setting *Ordering Key* = `${session}` on `PublishPulsarRecord` gets no validation error, no log
line, and no ordering key on any message. This mirrors the pre-existing *Message Key* wart, but adds
a second instance of it. Either scope `ORDERING_KEY` to `PublishPulsar` or reject it in
`customValidate` when *Ordering Key Field* is the operative property.

### 3. Blank handling diverges between the two paths (low)

`PublisherLease.java:683` uses `asString.isEmpty()`; `AbstractPulsarProducerProcessor.getOrderingKey`
(line 523) uses `StringUtils.isBlank`. So `"  "` means "no ordering key" on `PublishPulsar` and a real
whitespace ordering key on `PublishPulsarRecord`. A CSV/JSON source with padded-but-empty fields
silently creates a `" "` ordering-key bucket. Low impact, but the two are documented as a matched
pair ("null or empty" vs "blank").

### 4. The IT's control assertion is probabilistic (low)

`PublishPulsarOrderingKeyIT.java:149`. `spread.first > 0 && spread.second > 0` relies on
`ConsistentHashingStickyKeyConsumerSelector` splitting 40 fixed keys across two consumers whose ring
positions derive from auto-generated random consumer names. Very likely, not guaranteed — so this can
fail spuriously. Separately, `publishToTwoKeySharedConsumers` exits its drain on the 60 s deadline as
well as on count, and the control run never asserts all 40 arrived, so a slow broker yields a passing
control from a partial sample. Pinning explicit consumer names, or asserting
`spread.first + spread.second == MESSAGES` for the control too, makes it deterministic.

---

Only finding 1 is behaviour-affecting; the rest are tidy-ups. Nice work on keeping the null path
identical — that is what makes this safe to ship alongside the other two fixes.

jselvareef:
Thanks — all four in `8a064c3`, PR body revised. One thing I found on the way that you should know about.

**1. `Byte[]` / `Record`.** The ordering key field now goes through the same helper as the message key field (`keyBytes`), so both properties give the same bytes for the same value: `byte[]` as is, boxed bytes unboxed, a nested `Record` serialised with the FlowFile's writer, else UTF-8 text. The new test reads Avro through a real `AvroReader` with a `bytes` field named under both properties, and expects `s-1, s-1, s-2` as ordering keys and the same as message keys.

The finding along the way: on this NiFi line `AvroTypeUtil` hands an Avro `bytes` field over as an **`Object[]` whose elements are `Byte`**, not as a `Byte[]`. So the existing `instanceof Byte[]` branch in `getMessageKey` never fired for Avro input either — the first run of the new test showed *both* keys as `[Ljava.lang.Object;@5cf57368`. Mirroring the branch alone would not have fixed the ordering key, and the message key had the same defect. The shared helper now accepts either shape (`Byte[]`, or an `Object[]` whose every element is a `Byte`). That means this PR also corrects *Message Key Field* on an Avro `bytes` field, which is outside #196's scope; I've called it out in the PR body's "Not done" and am happy to split it into its own issue and PR if you would rather keep this one to the ordering key.

**2. *Ordering Key* on `PublishPulsarRecord`.** The descriptor and `getOrderingKey(context, flowFile)` moved to `PublishPulsar`, which appends it to the base list; `PublishPulsarRecord` no longer offers it. A unit test pins which processor offers which property. I went with scoping rather than a validation error because a property that is never read should not be in the UI at all; *Message Key* still has the pre-existing version of this and I left it alone.

**3. Blank.** A blank string in the ordering key field is "no ordering key", matching `StringUtils.isBlank` on `PublishPulsar`'s property; tested. The message key's treatment of blank text is unchanged.

**4. IT determinism.** The two `Key_Shared` consumers are now named `first` and `second`, so the hash ring — and the split of the 40 fixed keys — is the same on every run, and the control asserts that all 40 arrived before it judges the spread. Three consecutive runs, 4/4 each.

Verification on the head: 360/360 unit; `PublishPulsarOrderingKeyIT` ×3, `PublishPulsarMessageKeyIT`, `PublishPulsarRecordSchemaIT` against a real broker, all green.

david-streamlio:
All four look good, and the `Object[]` finding is the important one here — thank you for chasing it
rather than taking my suggestion at face value.

My review said "mirror the existing `Byte[]` branch". **That would not have worked**, and you only
know that because you ran it: `AvroTypeUtil` on this NiFi line hands an Avro `bytes` field over as an
`Object[]` whose elements are `Byte`, so the `instanceof Byte[]` branch never fired for Avro input
either, and both keys came out as `[Ljava.lang.Object;@…`. My finding named the right symptom and
prescribed a fix that would have left it in place. Accepting either shape is the right call.

**On splitting the *Message Key Field* fix: keep it here.**

Normally I would say split — it is outside #196's scope and the PR body is right to flag it. But the
two properties now go through one helper, which is the whole reason the ordering key is correct, and
separating them would mean either duplicating the shape handling or landing a helper that one caller
deliberately does not use. Neither is better than a slightly wider PR.

What I would ask for instead:

- **File the issue anyway**, closed by this PR, so the *Message Key Field* fix is findable from the
  issue tracker by someone who hits `[Ljava.lang.Object;@…` in a flow and searches for it. The PR
  body is the wrong place for that to live on its own.
- **Call it out in the release notes as a fix in its own right**, not as a footnote to the ordering
  key. Anyone using *Message Key Field* with an Avro `bytes` field has been getting a per-message
  identity hash — that is a silent partitioning bug, and it is older than #196.

Worth noting how far back it goes: `getMessageKey`'s `Byte[]` branch carries a comment explaining
exactly why it exists, so someone knew about the boxing and wrote a branch for the shape they
expected. It has presumably never fired on Avro input since. A test with a real `AvroReader`, which
is what you added, is the only thing that would have caught it — the branch reads correctly.

Scoping *Ordering Key* to `PublishPulsar` rather than validating it away is the better call too, for
the reason you give: a property that is never read should not be in the UI. Leaving the pre-existing
*Message Key* instance alone is right for this PR — worth its own issue whenever someone touches that
path.

jselvareef:
Done, in `c003856`:

- **#226** filed for the *Message Key Field* / Avro `bytes` defect, with the `AvroTypeUtil.convertByteArray` shape (`Object[]` of `Byte`, checked on the 2.11.0 and 2.9.0 artifacts), the Mockito output, the scope, and a pointer here; this PR now says `Closes #226` as well.
- **Release notes**: `docs/release-notes/2.11.0.1.md` carries the ordering key under *New* and #226 under *Fixes* as its own entry — what the key was, what it did to routing and compaction, and what a keyed flow should expect after upgrading — plus an *Upgrading* paragraph. The README's *Publisher message metadata* section gets a *Behaviour change since `2.11.0`* callout for the same reason: a flow keying by an Avro `bytes` field will start landing on the partition the key hashes to.

One mechanical note: #220 creates the same release-notes file for its own entry. Whichever of the two merges second will conflict on it; I'll rebase that one and fold the sections together as soon as the first lands, so it should not cost you anything beyond a second look.

On the `Byte[]` branch's history — agreed, and I would add that `MockRecordParser` yields every field as a `String`, so no unit test in the module could ever have reached either branch; the `AvroReader`-based test is now the one that does. The pre-existing *Message Key* on `PublishPulsarRecord` I have left as it is, per your note.

nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/utils/PublisherLease.java
```diff
@@ -630,37 +713,78 @@ protected CompletableFuture<MessageId> send(Producer producer, String key, Map<S
...
+            failed.completeExceptionally(e);
+            return failed;
+        }
+    }
+
     private String getMessageKey(final FlowFile flowFile, final RecordSetWriterFactory writerFactory,
                                  final Object keyValue) throws IOException, SchemaNotFoundException {
-        final byte[] messageKey;
+        final byte[] messageKey = keyBytes(flowFile, writerFactory, keyValue);
+        return (messageKey == null) ? null : new String(messageKey);
```
david-streamlio: **Medium — binary key bytes are decoded with the platform default charset, and this PR is what makes the line reachable.**

`new String(messageKey)` has no charset argument, so it uses `Charset.defaultCharset()`. Before this PR an Avro `bytes` field fell through to `toString()` and never reached here; the #226 fix now routes it through `keyBytes()` straight into this decode.

Concrete failure: an Avro `bytes` field holding a raw 16-byte UUID, named as *Message Key Field*. Under UTF-8 every invalid byte sequence decodes to U+FFFD, so two different UUIDs can produce the **identical** message key. They then route to the same partition, and on a compacted topic one supersedes the other — data loss. Before this PR each UUID got a unique-if-ugly key, so nothing was ever dropped. The result is also `file.encoding`-dependent, so two NiFi nodes in one cluster can derive different keys from the same record.

This also makes the parity claim in the README and in `keyBytes`'s javadoc inaccurate — naming one field under both properties does *not* currently give the same key twice: the ordering key gets the raw bytes via `orderingKey(byte[])`, the message key gets this lossy decode. Note `PublishPulsarOrderingKeyTest`'s Avro assertion passes only because `s-1`/`s-2` are ASCII.

Minimum fix: `new String(messageKey, StandardCharsets.UTF_8)` — deterministic across nodes, though still lossy for non-text bytes. Better: send byte-valued key fields through `keyBytes()` on the message key too, the way the KeyValue `SEPARATED` path already does, or reject a byte-typed *Message Key Field* with a clear validation error. Whichever you pick, the docs' "same key twice" sentence needs to match.

nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/utils/PublisherLease.java
```diff
@@ -630,37 +713,78 @@ protected CompletableFuture<MessageId> send(Producer producer, String key, Map<S
...
         } else {
-            final String keyString = keyValue.toString();
-            messageKey = keyString.getBytes(StandardCharsets.UTF_8);
+            return keyValue.toString().getBytes(StandardCharsets.UTF_8);
         }
-        return (messageKey == null) ? null : new String(messageKey);
+    }
+
+    /** True for an array whose every element is a {@link Byte} - a byte string in the Record API's clothing. */
+    private static boolean isBoxedBytes(final Object[] array) {
```
david-streamlio, line 811: **Low — `isBoxedBytes` returns `true` for any empty array.**

For an empty `Object[]`, `array instanceof Byte[]` is `false`, the `for` body never executes, and control reaches `return true`. So `new String[0]`, `new Integer[0]` and `new Object[0]` all classify as boxed bytes.

This is reachable: NiFi hands every Avro `array<T>` field to `record.getValue()` as an `Object[]` (same shape `writeAsJson`'s ARRAY branch handles a few lines up). Name an `array<string>` field as *Message Key Field* and you get a split behaviour — records whose array is empty take this branch and get key `""`, records whose array is non-empty fail the element check and fall through to `toString()`. All the empty-array records collapse onto one shared key and supersede each other on a compacted topic, where previously each had a distinct (identity-hash) key.

```java
private static boolean isBoxedBytes(final Object[] array) {
    if (array.length == 0) {
        return array instanceof Byte[];
    }
    ...
}
```

That keeps a genuinely empty `Byte[]` (an empty byte string) classified as bytes while an empty `String[]` is not.

nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/utils/PublisherLease.java
```diff
@@ -288,6 +311,8 @@ public void publish(final FlowFile flowFile, final RecordSet recordSet, final Re
 
                 final byte[] messageContent;
                 final String messageKey;
+                final byte[] orderingKey = orderingKeyField == null || orderingKeyField.isEmpty()
```
david-streamlio, line 319: **Low — a misspelled *Ordering Key Field* fails completely silently.**

`record.getValue(orderingKeyField)` returns `null` for a field that is not in the schema at all, which is indistinguishable here from a field that is present and null. So a typo makes `orderingKey` `null` for every record, and Pulsar silently falls back to the message key.

Concrete: someone types `sessionId` for a field actually named `session_id`. `Key_Shared` dispatch keeps following the message key, the ordering guarantee the flow was built for never exists, and there is no error, no warning, no `failure` routing — the flow looks perfectly healthy. The only way to notice is to observe dispatch on a multi-consumer subscription.

A one-time `logger.warn` when `record.getSchema().getField(orderingKeyField)` is empty would close it. *Message Key Field* has the same gap today so it is not a regression, but it bites harder for the ordering key: a wrong message key is at least visible on the message, whereas the ordering key's only observable effect is dispatch.

nifi-pulsar-processors/src/test/java/org/apache/nifi/processors/pulsar/it/PublishPulsarOrderingKeyIT.java
```diff
@@ -0,0 +1,276 @@
...
+    public void keySharedDispatchFollowsTheOrderingKeyNotTheMessageKey() throws Exception {
+        // control: no ordering key, distinct message keys spread over both consumers
+        final String controlTopic = topic("control");
+        final TestRunner control = publishPulsar();
+        control.setProperty(AbstractPulsarProducerProcessor.TOPIC, controlTopic);
+        control.setProperty(AbstractPulsarProducerProcessor.MESSAGE_KEY, "${key}");
+        final Distribution spread = publishToTwoKeySharedConsumers(control, controlTopic, false);
+        assertEquals("control: every message must have arrived", MESSAGES, spread.first + spread.second);
+        assertTrue("control: " + MESSAGES + " distinct message keys with no ordering key should reach both "
+                + "consumers, but landed " + spread, spread.first > 0 && spread.second > 0);
```
david-streamlio, line 160: **Low — the control run depends on broker hash-ring internals, and one assertion order hides its own diagnostic.**

The control is the right idea and it does rule out a broker that ignored the ordering key — that is the part of this IT only a broker can answer, and it is worth having. But `spread.first > 0 && spread.second > 0` rests on Pulsar's `Key_Shared` AUTO_SPLIT consistent-hash ring: its point count, its hash function, and the classic-vs-new sticky-key implementation whose default has shifted within the 4.x line. Fixing the consumer names makes the split reproducible *on a given broker*, not across brokers. Since `pulsar.image` is a property meant to track the client version, a future bump can land all 40 `device-N` keys in one consumer's range and fail the control with no product regression behind it.

`KeySharedPolicy.stickyHashRange()` with an explicit range per consumer would make both the control and the dispatch run deterministic and independent of broker defaults, and would keep the control's meaning intact.

Separately, in the dispatch run `assertTrue(together.first == 0 || together.second == 0)` is asserted *before* the arrival-total assertion. If the drain times out and only some messages arrive, the single-consumer assert passes vacuously and the failure surfaces one line later as a count mismatch — a misleading message for what is actually a delivery problem. Swapping the two lines gives the accurate diagnostic, the same ordering the control run already uses.

david-streamlio (commented):
Second review pass, high effort, against `74bd8e0` in a clean worktree. `mvn -pl nifi-pulsar-processors test`: 378/378 green, including the 11 new `PublishPulsarOrderingKeyTest` cases.

**The design holds up.** The `orderingKey == null` delegation in all four `send*` variants preserves the overridable-method contract the test suite relies on, the KeyValue `SEPARATED` path threads the key correctly and keeps the *Message Key Field* conflict check, and scoping *Ordering Key* to `PublishPulsar` via the `PROPERTIES = base + ORDERING_KEY` override is right and asserted. I also re-confirmed the #226 premise independently: `AvroTypeUtil.convertByteArray` hands back an `Object[]` of `Byte`, not a `Byte[]`, so the old `instanceof Byte[]` branch never matched an `AvroReader` `bytes` field. Your `Object[]` finding was the correct call.

Four findings inline. **The first one I'd want fixed before this merges** — it is a data-loss path that this PR opens, not a pre-existing one, precisely *because* the #226 fix now routes Avro `bytes` fields through a line they never used to reach. The other three are low severity.

Two housekeeping items outside the code:

- The PR body's *What changed → Docs* bullet still says "New `docs/release-notes/2.11.0.1.md`" — stale from the second commit. The file is `2.11.0.2.md`, and the *Not done* section already says so correctly. `2.11.0.1.md` is indeed untouched.
- `docs/release-notes/2.11.0.2.md` collides with #228, which creates the same new file. You flagged this in *Not done*; noting that both PRs also touch `README.md`, and #217 touches it as well, so whichever lands last has three-way README context to fold.

jselvareef:
All four in `47c1649`, plus the two housekeeping items. PR body revised.

**1. Binary key bytes.** Taken the "better" option: a binary key field — `byte[]`, or the boxed bytes an `AvroReader` yields — is sent through `keyBytes()`, the way the KeyValue `SEPARATED` path already does, so it travels base64-encoded and flagged and two different byte strings are always two different keys. `getMessageKey(FlowFile, writerFactory, Object)` and its `new String(bytes)` are gone; a text key (text, number, or a nested record the writer serialised) is read as UTF-8, so every node derives the same key. Unit test with the pair you describe in spirit — `FF FE 01` and `FE FF 01`, asserted to be the same text under UTF-8 so the test cannot pass by accident — expecting two `keyBytes` calls and no `key(String)`; and an IT: the consumer sees `hasBase64EncodedKey()`, `getKeyBytes()` equal to the field, and two distinct keys. The parity wording in README, `additionalDetails` and the `keyBytes` javadoc now says what is actually true: the same field yields the same *bytes* under both properties, sent as a binary key on one side and the raw ordering key on the other; "same key twice" only for text.

**2. `isBoxedBytes` on an empty array.** As you wrote it: `Byte[]` is bytes, any other empty array is not. Test: two records whose `array<string>` field is empty, named as *Message Key Field*, produce neither `keyBytes` nor `key("")`.

**3. Misspelt field.** One warning per FlowFile, at the top of `publish`, when the record set's schema has no field of that name — naming the property, the field and the schema's fields. Applied to *Message Key Field* as well as *Ordering Key Field*: same line of code, same silence, and you noted the gap was already there. Three tests: each property misspelt warns exactly once and sets no key; a present field warns nothing. One honest caveat in the PR body: a reader that keeps unknown fields can still supply a value for a name the schema lacks, so the warning can be a false alarm there; its text ("not a field of the records' schema") stays accurate.

**4. IT.** The two `Key_Shared` consumers now hold explicit sticky hash ranges (`0–32767`, `32768–65535`), so a key's consumer is `hash(key) % 65536` against those ranges on any broker rather than a property of the AUTO_SPLIT ring; the control keeps its meaning. The dispatch run asserts arrival before the single-consumer check, as the control already did.

Housekeeping: the stale `2.11.0.1.md` mention in the body is fixed, and the README three-way with #228 and #217 is noted — I'll fold whichever lands last.

Verification: 383/383 unit on top of `main` after `2.11.0.1`; `PublishPulsarOrderingKeyIT` 5/5, `PublishPulsarMessageKeyIT` 4/4, `PublishPulsarRecordSchemaIT` 3/3, `KeyValueSchemaRoundTripIT` 7/7 against a real broker. Fails-first for the five new unit cases against the previous commit is in the body.

david-streamlio:
Merging this. The charset fix is the right one — routing a binary key field through `keyBytes()` rather than making the decode deterministic means two distinct byte strings stay two distinct keys, which is the property that actually matters on a compacted topic. I checked `isBoxedBytes` and the text/binary split at line 380 against the head; both are as described.

One trivial thing to sweep up, not worth holding the merge for:

`getOrderingKey`'s javadoc still reads *"converted exactly as `{@link #getMessageKey}` converts the message key field"* (`PublisherLease.java:693`), but `getMessageKey` was removed in `47c1649` — there is no method by that name left in the class. So the link is dangling, and it is in the one comment that documents the parity contract between the two properties. Either point it at `keyBytes` or restate it in prose.

Happy for that to ride along with whatever touches the file next rather than becoming its own PR.

Issue #53, opened 2023-03-02
Memory leak in consumer in asynchronous mode


Nkeramov:
Hello, recently I encountered a memory leak issue in the consumer's processor in asynchronous mode.

I use Java 11, Apache NiFi 1.20.0, Apache Pulsar 2.10.3. 

I builded pulsar bundle for Nifi with following depency versions
```
 <commons-collections4.version>4.4</commons-collections4.version>

 <commons-compress.version>1.21</commons-compress.version>

 <conscrypt-openjdk.version>2.5.2</conscrypt-openjdk.version>

 <groovy-eclipse-batch.version>2.5.6-01</groovy-eclipse-batch.version>

 <grpc.version>1.41.0</grpc.version>

 <jackson-core.version>2.13.4</jackson-core.version>

 <junit.version>4.13.2</junit.version>

 <mockito-core.version>3.12.4</mockito-core.version>

 <protobuf3.version>3.20.3</protobuf3.version>

 <protoc.version>3.20.0</protoc.version>

 <slf4j-simple.version>2.0.0</slf4j-simple.version>

 <nifi.version>1.20.0</nifi.version>

 <pulsar.version>2.10.3</pulsar.version>

 <java.version>11</java.version>

 <maven-compiler-plugin.version>3.10.1</maven-compiler-plugin.version>

 <maven-gpg-plugin.version>3.0.1</maven-gpg-plugin.version>

 <maven-javadoc-plugin.version>3.3.1</maven-javadoc-plugin.version>

 <maven-source-plugin.version>2.2.1</maven-source-plugin.version>

 <nifi-nar-maven-plugin.version>1.3.5</nifi-nar-maven-plugin.version>

 <nexus-staging-maven-plugin.version>1.6.13</nexus-staging-maven-plugin.version>

 <maven-versions-plugin.version>2.5</maven-versions-plugin.version>
```

I send small data packet with data from OpenWeatherMap site. I use ConsumePulsar and ProducePulsar in asynchronous mode. Data is written and read successfully, without loss. However, after a while, the heap size increases dramatically.

![Screenshot from 2023-02-27 13-54-58](https://user-images.githubusercontent.com/90603794/222432873-0f86cbc8-5601-40d7-9fcc-f3b114c15117.png)

I also builded this bundle for Apache NiFi 1.18.0 and tested my pipeline in it. The memory problem persisted

Here my pipeline in NiFi

[...]

PR #228, opened 2026-09-08
fix: stop truncating the consumer time properties to whole seconds (#225)

### LINKED ISSUE #225: Consumer time properties silently truncate fractional seconds (Auto Update Partition Interval, Expire Time of Incomplete Chunked Message)
Two consumer time properties are handed to the Pulsar client as **whole seconds**, so any fraction a
user types is silently discarded — the flow runs on a different value from the one configured, and
nothing says so.

```java
.autoUpdatePartitionsInterval(context.getProperty(AUTO_UPDATE_PARTITION_INTERVAL)
        .asTimePeriod(TimeUnit.SECONDS).intValue(), TimeUnit.SECONDS)
.expireTimeOfIncompleteChunkedMessage(context.getProperty(EXPIRE_TIME_OF_INCOMPLETE_CHUNKED_MESSAGE)
        .asTimePeriod(TimeUnit.SECONDS), TimeUnit.SECONDS)
```

`asTimePeriod(SECONDS)` truncates. Both properties use `TIME_PERIOD_VALIDATOR`, which happily accepts
a millisecond value, so:

| Configured | Actually applied |
|---|---|
| `90500 millis` | 90 s |
| `1500 millis` | 1 s |
| `500 millis` | 0 s |

The last row is the interesting one: `autoUpdatePartitionsInterval(0, SECONDS)` is not "as often as
possible", it is whatever the client does with zero — and either way it is not what was asked for.

### Why this is being filed now

#217 added *Topics Pattern Discovery Interval*, which has exactly the same shape, and its validator
rejects any value that is not a whole number of seconds. That makes the new property **stricter than
its two neighbours**: the same mistake is fatal on one and invisible on the others, two lines apart in
the same builder chain.

The inconsistency is real, and the right direction is to tighten these two rather than loosen the new
one — silently running on a value the user did not choose is the worse behaviour.

### Why it is not part of #217

Tightening these would **invalidate flows that run today**. Anyone who set `90500 millis`, or any
other fractional value, currently has a working processor that would stop being valid on upgrade.
That deserves its own change, its own release note, and a decision about whether it is a validation
error or an `@OnScheduled` warning — the latter is what #220 chose for a comparable case, and what
#217 chose for its own Read Compacted / Latest combination.

### Options

- **Reject** non-whole-second values, matching *Topics Pattern Discovery Interval*. Consistent and
  honest; breaks existing flows on upgrade.
- **Warn at `@OnScheduled`**, naming the configured and applied values. Keeps every flow valid and
  puts the truth where someone sees it once per start.
- **Round up rather than truncate**, so a sub-second value becomes 1 s instead of 0. Removes the worst
  case without touching validity, but still silently changes the value.

My preference is the warning, with the sub-second-rounds-to-zero case called out specifically, since
that is the one where the applied value is not merely coarser but meaningless.

Refs #217.

jselvareef:
Closes #225.

Two consumer time properties reached the Pulsar client as whole seconds, so any fraction the user typed was dropped in silence. Reading the client (4.2.4) showed the two are not the same case.

> **Revised after review** (third commit): one rule for both properties, the one #217 applies to *Topics Pattern Discovery Interval* — a value must be representable in the granularity the client stores it in, or it is rejected at validation. *Auto Update Partition Interval*: whole seconds, 1 to `Integer.MAX_VALUE` (the `intValue()` wrap past that range is closed). *Expire Time of Incomplete Chunked Message*: whole milliseconds, `0` kept as the documented "never expire". The `@OnScheduled` warning is gone, since a rejected value never reaches it. README callout re-scoped from `2.11.0` to every release since `2.1.0`. Details in the review thread.

### What changed

- **`AbstractPulsarConsumerProcessor.getConsumerBuilder`** — *Expire Time of Incomplete Chunked Message* is passed as `asTimePeriod(MILLISECONDS), MILLISECONDS`. `ConsumerBuilder.expireTimeOfIncompleteChunkedMessage(long, TimeUnit)` stores `unit.toMillis(duration)`, so the whole-seconds conversion was entirely ours and every configured value is now applied exactly. What the truncated `0` did: `ConsumerImpl` schedules the chunk-expiry task only when the value is `> 0`, so a flow that set `500 millis` has been running with expiry **disabled**, not fast.
- **`customValidate`** — *Auto Update Partition Interval* must be a whole number of seconds between 1 and `Integer.MAX_VALUE`. The client keeps this interval as an `int` of whole seconds (`ConsumerConfigurationData.setAutoUpdatePartitionsIntervalSeconds` → `unit.toSeconds`) and begins with `checkArgument(interval > 0, "interval needs to be > 0")`, so: `asTimePeriod(SECONDS).intValue()` handed it `0` for a sub-second value and the builder threw on every trigger (see the probe below); a fractional value such as `90500 millis` ran as 90 s while the configuration said otherwise; and a value past the `int` range wrapped — `30000 days` to a negative interval the client refused, `10000 weeks` to a positive one about 55 years long instead of 191. One condition rejects all three. *Expire Time of Incomplete Chunked Message* must be a whole number of milliseconds, since that is the `long` the client keeps; `0.5 millis` would otherwise truncate to the `0` the client reads as "never expire", the opposite of what was asked. `0` itself stays valid as the documented sentinel.
- **No `@OnScheduled` warning.** The first version warned about a fractional interval at scheduling; under the rule above a fractional value is invalid, and a processor that fails validation is never scheduled, so the warning could not fire. The validation message carries the explanation: the property, the value, and what the client can represent.
- **Descriptions** of both properties say what the client keeps and what is therefore valid.
- **README** — new *Consumer time properties* section stating the rule and the two granularities, with a *Behaviour change since `2.1.0`* callout: the whole-seconds conversion of the chunk expiry dates from `4ed35de` (2024) and is in every release, so an operator coming from `2.10.0.1` is exactly who it addresses.

Passing milliseconds to `autoUpdatePartitionsInterval` was considered and rejected: `500` passes the `> 0` check, `toSeconds(500)` is `0`, and the partition-update timer would then fire continuously — worse than the exception it replaces.

### Tests

**`ConsumePulsarTimePropertiesTest`** (new, 13 cases) builds the consumer configuration through a real `PulsarClient` (`pulsar://localhost:1`; building a client opens no connection and nothing here subscribes) and asserts against the `ConsumerConfigurationData` the builder holds — the same idea as `PublishPulsarProducerConfigurationTest`: what the client actually ends up with, not what the processor thinks it passed.

- chunk expiry keeps its fraction (`1500 millis` → 1500 ms), a sub-second value is applied rather than disabling expiry (`500 millis` → 500 ms), the default is one minute;
- a partition update interval under a second is rejected, exactly one second is valid; `90500 millis` is rejected; `30000 days`, `10000 weeks` and `Integer.MAX_VALUE + 1` seconds are rejected; `Integer.MAX_VALUE` seconds is valid and applied; `90 sec` is applied as given with no warning;
- the client refuses a zero-second interval (pins the reason for the floor, so a client that relaxes it is noticed);
- a chunk expiry of `0.5 millis` or `1500000 nanos` is rejected; `0 sec` is valid and applied as 0.

`TestConsumePulsar.chunkedMessageConfigTest` verified the old `expireTimeOfIncompleteChunkedMessage(120, SECONDS)` call and now verifies `(120_000, MILLISECONDS)`.

### Fails first

The new test class against `main` (`00a6eb8`) — 4 of 8 fail:

```
Tests run: 8, Failures: 4, Errors: 0, Skipped: 0 -- in ConsumePulsarTimePropertiesTest
  theChunkExpiryKeepsItsFractionOfASecond:                        expected:<1500> but was:<1000>
  aSubSecondChunkExpiryIsAppliedRatherThanDisablingExpiry:        expected:<500> but was:<0>
  aPartitionUpdateIntervalUnderOneSecondIsRejected:               Processor appears to be valid but expected it to be invalid
  aFractionalPartitionUpdateIntervalIsAppliedAsWholeSecondsAndWarnsWhenScheduled:
                                                                  one warning about the dropped fraction, got [] expected:<1> but was:<0>
```

And what a `500 millis` partition update interval did on `main`, from a throwaway probe (validation, then `getConsumerBuilder` with the real client):

```
PROBE validation results=[]
PROBE builder threw java.lang.IllegalArgumentException: interval needs to be > 0
```

Third commit, the new cases against the second commit's rules:

```
Tests run: 13, Failures: 4 -- in ConsumePulsarTimePropertiesTest
  aFractionalPartitionUpdateIntervalIsRejected:            Processor appears to be valid but expected it to be invalid
  aPartitionUpdateIntervalPastTheIntRangeIsRejected:       Processor appears to be valid but expected it to be invalid
  aChunkExpiryUnderOneMillisecondIsRejected:               Processor appears to be valid but expected it to be invalid
  aChunkExpiryWithAFractionOfAMillisecondIsRejected:       Processor appears to be valid but expected it to be invalid
```

### Verification

- `mvn test` on the current head: **381/381** (on top of `main` after `2.11.0.1`: 367 + 13 + the one updated verification).
- Third commit: `ConsumePulsarIT` (4) and `ConsumePulsarTopicSchemaIT` (6) against a real broker.
- `ConsumePulsarIT` (4), `ConsumePulsarRestartIT` (6), `ConsumePulsarTopicSchemaIT` (6) against a real broker (Testcontainers, `apachepulsar/pulsar:4.2.4`): all pass; the consumer configuration the client logs shows `expireTimeOfIncompleteChunkedMessageMillis: 60000` for the default, as before.

### Not done

- ~~No release-notes entry in this PR~~ — added in the second commit: since `2.11.0.1` has shipped, the entry goes in a new `docs/release-notes/2.11.0.2.md` (*Behaviour changes*: chunk expiry honoured to the millisecond; sub-second partition update interval rejected, with the note that such flows were failing at run time; plus *Upgrading*). #222 creates the same file for its own entries; whichever lands second I rebase and fold together. Rebased onto `main` after `2.11.0.1`; the only code conflict was the two `@OnScheduled` helpers landing on the same lines, both kept.
- The property descriptions are the only per-processor documentation for these two; neither had an `additionalDetails` paragraph before and I did not add one — the README section covers the reasoning.
- The producer-side *Auto Update Partition Interval* has the same truncation and, because it goes through `loadConf()`, skips the client's zero guard entirely; filed as #230, not touched here.
- `Ack Timeout` and *Negative Acknowledgment Redelivery Delay* are passed in milliseconds and microseconds respectively and do not truncate; I checked and left them alone. *Topics Pattern Discovery Interval* is #217's and already rejects a fraction.
- The new test builds a real `PulsarClient` per case (about 2 s each, ~17 s for the class); a shared client would be faster but would carry state between cases.

https://claude.ai/code/session_016FxYKxcaKpvRbJSYP3QYZ9

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,26 @@ on a `Key_Shared` subscription**: a consumer receives a whole batch at a time, s
 several keys hands one consumer messages belonging to another consumer's key range. It has no
 effect when batching is off.
 
+## Consumer time properties
+
+Two consumer properties are handed to the Pulsar client in units coarser than NiFi lets you type
+them in, and behave accordingly.
+
+*Auto Update Partition Interval* is kept by the client in **whole seconds**, and the client refuses
+zero. A value under one second is therefore rejected at validation — before this rule it validated
+and then failed on every trigger with the client's `interval needs to be > 0`, about an interval
+nobody typed. A value of a second or more with a fraction (`90500 millis`) runs, on the whole
+seconds the client keeps (`90`), and the processor logs a warning saying so when it starts.
+
+*Expire Time of Incomplete Chunked Message* is kept by the client in milliseconds and is applied
+exactly as configured.
+
+> **Behaviour change since `2.11.0`:** *Expire Time of Incomplete Chunked Message* used to be
+> converted to whole seconds on the way to the client, so a fraction was dropped and a sub-second
+> value became `0` — which the client reads as **never expire**: incomplete chunks were kept until
+> the pending-chunk queue evicted them. A flow that set `500 millis` now expires them after 500 ms,
+> and `1500 millis` means 1.5 s rather than 1 s. Whole-second values are unchanged.
+
 ## Consuming from topics that have a schema
 
 `ConsumePulsarRecord`'s **Message Schema Strategy** decides how a message becomes records.
--- a/README.md
+++ b/README.md
@@ -231,23 +231,32 @@ effect when batching is off.
 
 ## Consumer time properties
 
-Two consumer properties are handed to the Pulsar client in units coarser than NiFi lets you type
-them in, and behave accordingly.
+NiFi time-period values always carry a unit (`500 millis`, `90 sec`, `1 min`), so the only question
+for a property the Pulsar client stores in a coarser unit is whether the duration you typed is
+representable in the granularity the client keeps. Where it is not, the value is **rejected at
+validation** rather than silently applied as something else.
 
-*Auto Update Partition Interval* is kept by the client in **whole seconds**, and the client refuses
-zero. A value under one second is therefore rejected at validation — before this rule it validated
-and then failed on every trigger with the client's `interval needs to be > 0`, about an interval
-nobody typed. A value of a second or more with a fraction (`90500 millis`) runs, on the whole
-seconds the client keeps (`90`), and the processor logs a warning saying so when it starts.
+*Auto Update Partition Interval* is kept by the client as a whole number of **seconds**, in an `int`,
+and the client refuses zero. The value must therefore be a whole number of seconds between `1 sec`
+and `2147483647 sec`: `500 millis` would reach the client as `0` and be refused on every trigger,
+`90500 millis` would run as `90 sec` while the configuration says otherwise, and anything past the
+`int` range wrapped around — `30000 days` became a negative interval and failed, `10000 weeks` ran
+silently as about 55 years instead of 191. *Topics Pattern Discovery Interval* follows the same rule
+for the same reason.
 
-*Expire Time of Incomplete Chunked Message* is kept by the client in milliseconds and is applied
-exactly as configured.
+*Expire Time of Incomplete Chunked Message* is kept by the client as a whole number of
+**milliseconds**, so any whole-millisecond value is applied exactly as configured; a fraction of a
+millisecond is rejected. `0` is a deliberate value: it disables the expiry, and incomplete chunks are
+then kept until the pending-chunk queue evicts them.
 
-> **Behaviour change since `2.11.0`:** *Expire Time of Incomplete Chunked Message* used to be
-> converted to whole seconds on the way to the client, so a fraction was dropped and a sub-second
-> value became `0` — which the client reads as **never expire**: incomplete chunks were kept until
-> the pending-chunk queue evicted them. A flow that set `500 millis` now expires them after 500 ms,
-> and `1500 millis` means 1.5 s rather than 1 s. Whole-second values are unchanged.
+> **Behaviour change since `2.1.0`:** every release so far converted *Expire Time of Incomplete
+> Chunked Message* to whole seconds on the way to the client, so a fraction of a second was dropped
+> and a sub-second value became `0` — which the client reads as **never expire**. A flow that set
+> `500 millis` had no chunk expiry at all; it now expires incomplete chunks after 500 ms, and
+> `1500 millis` means 1.5 s rather than 1 s. Whole-second values are unchanged. On *Auto Update
+> Partition Interval*, a value that is not a whole number of seconds in the `int` range is now
+> invalid; a sub-second value was already failing every trigger, so the only flows this stops are
+> ones that ran on a different interval from the one configured.
 
 ## Consuming from topics that have a schema
 
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
@@ -163,9 +163,9 @@ public abstract class AbstractPulsarConsumerProcessor<T> extends AbstractProcess
             .name("AUTO_UPDATE_PARTITION_INTERVAL")
             .displayName("Auto Update Partition Interval")
             .description("Set the interval of updating partitions (default: 1 minute). This only works if " +
-                    "autoUpdatePartitions is enabled. The Pulsar client keeps this interval in whole seconds: " +
-                    "the shortest interval it accepts is one second, and a fraction of a second is dropped - " +
-                    "the processor logs the value actually applied when it starts.")
+                    "autoUpdatePartitions is enabled. The Pulsar client keeps this interval as a whole number of " +
+                    "seconds, so the value must be a whole number of seconds between 1 second and 2147483647 seconds; " +
+                    "anything else is rejected at validation rather than silently applied as something different.")
             .addValidator(StandardValidators.TIME_PERIOD_VALIDATOR)
             .defaultValue("1 min")
             .required(false)
@@ -250,7 +250,8 @@ public abstract class AbstractPulsarConsumerProcessor<T> extends AbstractProcess
             .displayName("Expire Time of Incomplete Chunked Message")
             .description("If producer fails to publish all the chunks of a message then consumer can expire incomplete" +
                     " chunks if consumer won't be able to receive all chunks in expire times (default 1 minute). " +
-                    "Applied with millisecond precision.")
+                    "The Pulsar client keeps this as a whole number of milliseconds, so the value must be one; " +
+                    "0 disables the expiry, and incomplete chunks are then kept until the pending-chunk queue evicts them.")
             .addValidator(StandardValidators.TIME_PERIOD_VALIDATOR)
             .defaultValue("60 sec")
             .required(false)
@@ -465,17 +466,36 @@ public abstract class AbstractPulsarConsumerProcessor<T> extends AbstractProcess
                "Acknowledgment Timeout needs to be greater than 10 seconds.").build());
         }
 
-        // The client stores the partition update interval in whole seconds and refuses zero
-        // ("interval needs to be > 0"), so a value under a second reached it as 0 and the consumer could not
-        // be created: every trigger failed with a message about an interval the user never typed (#225).
-        // Rejecting it here fails nothing that runs today.
+        // Two time properties are handed to the client in a coarser unit than NiFi lets the user type, and
+        // the rule for each is the same: the value must be representable in the granularity the client
+        // stores it in, or it is rejected - never silently applied as something else (#225).
+        //
+        // The partition update interval is an int of whole seconds on the client, which also refuses zero
+        // ("interval needs to be > 0"). So a value under a second reached it as 0 and the consumer could
+        // not be created; a fraction of a second was dropped; and a value past Integer.MAX_VALUE seconds
+        // wrapped in intValue() - negative, and refused, or positive and silently far shorter than asked.
         final long partitionUpdateIntervalMillis = validationContext.getProperty(AUTO_UPDATE_PARTITION_INTERVAL)
                 .asTimePeriod(TimeUnit.MILLISECONDS);
-        if (partitionUpdateIntervalMillis < TimeUnit.SECONDS.toMillis(1)) {
+        if (partitionUpdateIntervalMillis < TimeUnit.SECONDS.toMillis(1)
+                || partitionUpdateIntervalMillis % TimeUnit.SECONDS.toMillis(1) != 0
+                || TimeUnit.MILLISECONDS.toSeconds(partitionUpdateIntervalMillis) > Integer.MAX_VALUE) {
             results.add(new ValidationResult.Builder().valid(false).subject(AUTO_UPDATE_PARTITION_INTERVAL.getDisplayName())
-                .explanation("the Pulsar client keeps this interval in whole seconds and accepts nothing under one second; "
-                    + validationContext.getProperty(AUTO_UPDATE_PARTITION_INTERVAL).getValue()
-                    + " would reach it as 0 and the consumer could not be created")
+                .explanation("the Pulsar client keeps this interval as a whole number of seconds between 1 and "
+                    + Integer.MAX_VALUE + "; " + validationContext.getProperty(AUTO_UPDATE_PARTITION_INTERVAL).getValue()
+                    + " is not one, and would be applied as a different interval or refused by the client")
+                .build());
+        }
+
+        // The chunk expiry is a long of whole milliseconds on the client, and 0 means "never expire". A value
+        // between 0 and 1 ms would truncate to that 0 and disable the expiry for someone who asked for the
+        // shortest one; a fraction of a millisecond would be dropped.
+        final long expireTimeNanos = validationContext.getProperty(EXPIRE_TIME_OF_INCOMPLETE_CHUNKED_MESSAGE)
+                .asTimePeriod(TimeUnit.NANOSECONDS);
+        if (expireTimeNanos % TimeUnit.MILLISECONDS.toNanos(1) != 0) {
+            results.add(new ValidationResult.Builder().valid(false).subject(EXPIRE_TIME_OF_INCOMPLETE_CHUNKED_MESSAGE.getDisplayName())
+                .explanation("the Pulsar client keeps this as a whole number of milliseconds; "
+                    + validationContext.getProperty(EXPIRE_TIME_OF_INCOMPLETE_CHUNKED_MESSAGE).getValue()
+                    + " is not one (0 disables the expiry)")
                 .build());
         }
 
@@ -510,7 +530,6 @@ public abstract class AbstractPulsarConsumerProcessor<T> extends AbstractProcess
         // Exclusive subscription with "Exclusive consumer is already connected". The cache is built
         // lazily below and disposed in cleanUp().
         this.consumerCacheSize = context.getProperty(CONSUMER_CACHE_SIZE).asInteger();
-        warnIfPartitionUpdateIntervalLosesAFraction(context);
 
         if (context.getProperty(ASYNC_ENABLED).isSet() && context.getProperty(ASYNC_ENABLED).asBoolean()) {
             setConsumerPool(Executors.newFixedThreadPool(context.getProperty(MAX_ASYNC_REQUESTS).asInteger()));
@@ -537,21 +556,6 @@ public abstract class AbstractPulsarConsumerProcessor<T> extends AbstractProcess
                     + "no negative acknowledgement at all. Lower the delay unless the longer wait is intended.",
                     NEGATIVE_ACK_REDELIVERY_DELAY.getDisplayName(), context.getProperty(NEGATIVE_ACK_REDELIVERY_DELAY).getValue(),
                     ACK_TIMEOUT.getDisplayName(), context.getProperty(ACK_TIMEOUT).getValue());
-        }
-    }
-
-    /**
-     * The client keeps Auto Update Partition Interval in whole seconds, so a fraction of a second in the configured
-     * value is not applied. A value under a second is rejected by {@link #customValidate}; one of a second or more
-     * with a fraction runs, only coarser than asked, and this says so once per start rather than never (#225).
-     */
-    private void warnIfPartitionUpdateIntervalLosesAFraction(final ProcessContext context) {
-        final long configuredMillis = context.getProperty(AUTO_UPDATE_PARTITION_INTERVAL).asTimePeriod(TimeUnit.MILLISECONDS);
-        if (configuredMillis % TimeUnit.SECONDS.toMillis(1) != 0) {
-            getLogger().warn("{} is set to {}, which the Pulsar client applies as {} seconds: it keeps this interval in "
-                    + "whole seconds, so the fraction of a second is dropped",
-                    AUTO_UPDATE_PARTITION_INTERVAL.getDisplayName(), context.getProperty(AUTO_UPDATE_PARTITION_INTERVAL).getValue(),
-                    TimeUnit.MILLISECONDS.toSeconds(configuredMillis));
         }
     }
 
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
@@ -474,6 +474,13 @@ public abstract class AbstractPulsarConsumerProcessor<T> extends AbstractProcess
         // ("interval needs to be > 0"). So a value under a second reached it as 0 and the consumer could
         // not be created; a fraction of a second was dropped; and a value past Integer.MAX_VALUE seconds
         // wrapped in intValue() - negative, and refused, or positive and silently far shorter than asked.
+        //
+        // Unconditional on purpose, whatever Auto Update Partitions says: the client's precondition runs
+        // when the consumer is BUILT, for every consumer, while the value is only read at runtime under
+        // that flag. TIME_PERIOD_VALIDATOR already guarantees a non-negative duration, so the int overflow
+        // is the only way the builder can be handed a negative interval - which makes the upper bound a
+        // correctness requirement, not a sanity limit. The floor stays at one second, not zero, because
+        // that is what this setter enforces; the sibling Topics Pattern Discovery Interval accepts zero.
         final long partitionUpdateIntervalMillis = validationContext.getProperty(AUTO_UPDATE_PARTITION_INTERVAL)
                 .asTimePeriod(TimeUnit.MILLISECONDS);
         if (partitionUpdateIntervalMillis < TimeUnit.SECONDS.toMillis(1)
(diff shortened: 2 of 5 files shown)


nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
```diff
@@ -462,6 +465,20 @@ protected Collection<ValidationResult> customValidate(ValidationContext validati
                "Acknowledgment Timeout needs to be greater than 10 seconds.").build());
         }
 
+        // The client stores the partition update interval in whole seconds and refuses zero
+        // ("interval needs to be > 0"), so a value under a second reached it as 0 and the consumer could not
+        // be created: every trigger failed with a message about an interval the user never typed (#225).
+        // Rejecting it here fails nothing that runs today.
+        final long partitionUpdateIntervalMillis = validationContext.getProperty(AUTO_UPDATE_PARTITION_INTERVAL)
+                .asTimePeriod(TimeUnit.MILLISECONDS);
+        if (partitionUpdateIntervalMillis < TimeUnit.SECONDS.toMillis(1)) {
```
david-streamlio: **Medium — the floor guards only the lower bound, so the #225 failure mode survives at the top of the `int` range.**

Line 699 still reads `.asTimePeriod(TimeUnit.SECONDS).intValue()`, unchanged by this PR, and nothing bounds the value above. Verified arithmetic:

| Configured | Seconds | `.intValue()` |
|---|---|---|
| `30000 days` | 2,592,000,000 | **−1,702,967,296** |
| `10000 weeks` | 6,048,000,000 | **1,753,032,704** |

The first passes `assertValid()` and then throws `interval needs to be > 0` on every trigger — the exact symptom this PR exists to remove, just reached from the other end. The second is worse: it validates, never warns, and silently runs as ≈55 years instead of the ≈191 that was asked for. That is the silent-wrong-value class the PR is about.

**This is also where the whole-seconds rule belongs.** Rather than a floor plus a warning, I'd make this block enforce what the client can actually represent, matching what #217 already does for `Topics Pattern Discovery Interval`:

```java
final long millis = validationContext.getProperty(AUTO_UPDATE_PARTITION_INTERVAL)
        .asTimePeriod(TimeUnit.MILLISECONDS);
if (millis < 1000L || millis % 1000L != 0L
        || TimeUnit.MILLISECONDS.toSeconds(millis) > Integer.MAX_VALUE) { ... }
```

One rule, three problems: the sub-second case this PR already catches, the fractional case it currently only warns about, and the overflow above. See the summary for why I think rejecting the fraction is safe despite the compatibility argument — the short version is that sub-second values are already broken today, so the only flows affected are ones with a fractional-but-≥1s value, and they fail loudly and fixably rather than silently.

nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
```diff
@@ -668,8 +701,11 @@ protected synchronized ConsumerBuilder<GenericRecord> getConsumerBuilder(Process
                 .negativeAckRedeliveryDelay(context.getProperty(NEGATIVE_ACK_REDELIVERY_DELAY)
                         .asTimePeriod(TimeUnit.MICROSECONDS), TimeUnit.MICROSECONDS)
                 .autoAckOldestChunkedMessageOnQueueFull(context.getProperty(AUTO_ACK_OLDEST_CHUNKED_ON_QUEUE_FULL).asBoolean())
+                // The client stores this in milliseconds, so hand it over in milliseconds: converting to whole
+                // seconds here dropped any fraction, and turned a sub-second value into 0, which disables the
+                // expiry altogether (#225).
                 .expireTimeOfIncompleteChunkedMessage(context.getProperty(EXPIRE_TIME_OF_INCOMPLETE_CHUNKED_MESSAGE)
-                        .asTimePeriod(TimeUnit.SECONDS), TimeUnit.SECONDS)
+                        .asTimePeriod(TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS)
```
david-streamlio, line 719: **Low — this half of the fix is correct, but the trap moved rather than closed.**

Handing the client millis is exactly right: `unit.toMillis(v)` into a `long` millis field is lossless, so `500 millis` genuinely expires after 500 ms now.

What is missing is the matching floor. `Expire Time of Incomplete Chunked Message` got no validation, so "sub-second becomes 0 becomes never-expire" simply relocated from <1 s to <1 ms: `0.5 millis` passes `assertValid()`, `asTimePeriod(MILLISECONDS)` truncates it to `0`, and `ConsumerImpl.removeExpireIncompleteChunkedMessages()` gates `<= 0` as *expiry disabled*. So a user asking for the shortest possible expiry gets no expiry at all — silently.

That sits awkwardly with the property description this PR adds ("Applied with millisecond precision") and README line 244 ("applied exactly as configured"). And since the PR added a floor to the sibling property for precisely this reason, the asymmetry reads as unintentional rather than chosen.

A whole-millis floor is the counterpart of the seconds rule above: reject a value that is greater than zero but truncates to `0` ms. `0` itself is worth keeping as an explicit sentinel given the README already documents it as "never expire" — just make it deliberate rather than something you arrive at by typing `0.5 millis`.

nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
```diff
@@ -522,6 +540,21 @@ private void warnIfNegativeAckDelayExceedsAckTimeout(final ProcessContext contex
         }
     }
 
+    /**
+     * The client keeps Auto Update Partition Interval in whole seconds, so a fraction of a second in the configured
+     * value is not applied. A value under a second is rejected by {@link #customValidate}; one of a second or more
+     * with a fraction runs, only coarser than asked, and this says so once per start rather than never (#225).
+     */
+    private void warnIfPartitionUpdateIntervalLosesAFraction(final ProcessContext context) {
```
david-streamlio: **Low — the warning fires for a property that may have no effect, and its role changes under the whole-seconds rule.**

As written this is unconditional, so a processor with `Auto update partitions = false` (the default) and `1500 millis` logs a WARN at every start about a dropped fraction of an interval the client never uses. Gating it on `AUTO_UPDATE_PARTITIONS` would fix that. Note the *validation* rejection must stay unconditional — line 699 is called regardless of the flag, so an invalid value reaches the client either way.

If the whole-seconds rule lands, this method stops being the remedy for a fractional value and becomes something more useful: the thing that explains an existing `flow.xml.gz` value at the moment the processor goes invalid. Worth keeping for that, with the message reworded from "the fraction is dropped" to something that points at the new validation rule.

README.md
```diff
@@ -229,6 +229,26 @@ on a `Key_Shared` subscription**: a consumer receives a whole batch at a time, s
...
+*Auto Update Partition Interval* is kept by the client in **whole seconds**, and the client refuses
+zero. A value under one second is therefore rejected at validation — before this rule it validated
+and then failed on every trigger with the client's `interval needs to be > 0`, about an interval
+nobody typed. A value of a second or more with a fraction (`90500 millis`) runs, on the whole
+seconds the client keeps (`90`), and the processor logs a warning saying so when it starts.
+
+*Expire Time of Incomplete Chunked Message* is kept by the client in milliseconds and is applied
+exactly as configured.
+
+> **Behaviour change since `2.11.0`:** *Expire Time of Incomplete Chunked Message* used to be
```
david-streamlio: **Low but user-facing — the version in this callout is wrong, and wrong in the direction that makes operators skip it.**

`> **Behaviour change since 2.11.0:**` scopes the chunk-expiry change to 2.11.0, but the truncation was introduced in `4ed35de` ("Added chunked message properties", 2024-07-17). I checked the tags directly: the `asTimePeriod(TimeUnit.SECONDS)` conversion is present verbatim in `v2.9.0`, `v2.10.0` and `v2.11.0`.

This matters because the README uses that exact marker twice elsewhere (lines 110 and 216) for regressions genuinely introduced *in* 2.11.0. An operator upgrading from 2.10.0.1 with `Expire Time = 500 millis` reads "since 2.11.0", concludes it does not apply to them, and is then surprised when incomplete chunks start being dropped after upgrade — the one audience the callout exists for.

"Behaviour change since `2.1.0`" (or no version at all) is accurate. The release note itself does not make this mistake, only the README.

david-streamlio (commented):
High-effort review against `6f14763` in a clean worktree. **The unit analysis behind this fix is correct and I verified it independently** by decompiling `pulsar-client-4.2.4` rather than trusting the comments:

- `ConsumerBuilderImpl.expireTimeOfIncompleteChunkedMessage(long, TimeUnit)` does `unit.toMillis(v)` into `expireTimeOfIncompleteChunkedMessageMillis`, so handing it millis is exact — this half of the fix is a real precision gain.
- `ConsumerConfigurationData.setAutoUpdatePartitionsIntervalSeconds(int, TimeUnit)` really does `checkArgument(interval > 0, "interval needs to be > 0")`, so the PR's account is *more* accurate than #225's: the client refuses zero outright rather than doing something undefined with it.
- `ConsumerImpl.removeExpireIncompleteChunkedMessages()` opens with `if (expireTimeOfIncompleteChunkedMessageMillis <= 0) return;`, so "0 means never expire" in the README and release note is right.
- The test is not vacuous: 4 of its 8 cases fail against `origin/main` (`500 millis → 0`, `1500 millis → 1000`, the validation case, warn count 0), 8/8 pass here.

## The review position: make the rule per-property, and match #217

The substantive ask is one change of approach. This PR treats a fractional whole-second value as something to **warn about**; #217 treats the same input on `Topics Pattern Discovery Interval` as something to **reject**. Both properties are stored by the client in whole seconds, so they should answer the same question the same way — and #217's answer is the better one.

Worth being precise about the mechanism, because it makes the rule simpler than it looks. NiFi time-period values always carry their unit (`TIME_PERIOD_VALIDATOR`: "Must be of format `<duration> <TimeUnit>` … nanos, millis, secs, mins, hrs, days"), so nobody types a bare `1500` and there is no field-typing question to resolve. The only question per property is **whether the duration the user expressed is representable in the granularity the client stores it in.** That gives three rules:

| Property | Client granularity | Rule |
|---|---|---|
| Auto Update Partition Interval | seconds, `int` | whole seconds, ≥ 1 s, ≤ `Integer.MAX_VALUE` s |
| Topics Pattern Discovery Interval (#217) | seconds, `int` | whole seconds, ≥ 1 s, ≤ `Integer.MAX_VALUE` s |
| Expire Time of Incomplete Chunked Message | millis, `long` | whole millis, ≥ 1 ms (or 0 = never expire, documented) |

On the compatibility objection — this PR's rationale for warning instead of rejecting, and the reason #217's comment deferred to #225 — it is thinner than it looks. Sub-second values are **already** broken today (the client throws on every trigger), so rejecting them loses nothing that runs. Only a fractional-but-≥1s value like `90500 millis` is affected, the default is `1 min` so it had to be typed deliberately, and the resulting break is a processor that goes invalid on the canvas with a one-line fix and a message that explains itself — strictly better than a flow that silently runs at 90 s while its configuration says 90.5.

Keep `warnIfPartitionUpdateIntervalLosesAFraction`, though — not as the remedy, but because a `flow.xml.gz` that already holds a fractional value needs something to say so at the moment it goes invalid.

## Out of scope, worth its own issue

`AbstractPulsarProducerProcessor.java:436` — the identically named producer-side `AUTO_UPDATE_PARTITION_INTERVAL` keeps the old conversion with neither validator nor warning, and its failure is *worse* than the consumer's was. It goes `config.put("autoUpdatePartitionsIntervalSeconds", …)` → `loadConf()`, which Jackson maps onto the unchecked single-arg setter, **bypassing the client's `interval needs to be > 0` guard entirely** — so a sub-second value lands as `0 s` in silence instead of erroring. The new README section is honestly scoped "Consumer time properties", so nothing here is wrong; the same seconds rule applied there would close it.

Cleared as non-issues: neither property supports Expression Language, so `asTimePeriod` cannot meet an unresolved EL string; `customValidate` cannot throw on an unparseable value because NiFi runs property validators first; `ConsumePulsarRecord.customValidate` seeds from `super`, so the floor applies there too; and omitting the release note from the README compatibility table matches how 2.11.0.1 was handled.

README.md
```diff
@@ -229,6 +229,26 @@ on a `Key_Shared` subscription**: a consumer receives a whole batch at a time, s
 several keys hands one consumer messages belonging to another consumer's key range. It has no
 effect when batching is off.
 
+## Consumer time properties
+
+Two consumer properties are handed to the Pulsar client in units coarser than NiFi lets you type
```
david-streamlio: **Low — this sentence describes the state before the fix, and contradicts line 244.**

"Two consumer properties are handed to the Pulsar client in units coarser than NiFi lets you type them in, and behave accordingly" — but ten lines down the section says *Expire Time of Incomplete Chunked Message* "is kept by the client in milliseconds and is applied exactly as configured." After this PR only **one** of the two is coarser.

Saying so explicitly is also the more useful framing for the reader: one property is stored by the client in whole seconds and constrained accordingly, the other is stored in milliseconds and applied exactly. That is the distinction someone reading this section is trying to learn.

david-streamlio:
Follow-up to my review above, now that #217 is taking the same whole-seconds rule — the two properties converge, but **not quite identically**, and the difference is easy to get wrong when copying the rule across.

**1. The client's two preconditions are not the same.** I checked both in `pulsar-client-4.2.4`:

| Property | Setter precondition | So the floor is |
|---|---|---|
| *Auto Update Partition Interval* | `ConsumerConfigurationData.setAutoUpdatePartitionsIntervalSeconds`: `checkArgument(interval > 0)` | **≥ 1 s** — zero is refused outright |
| *Topics Pattern Discovery Interval* (#217) | `ConsumerBuilderImpl.patternAutoDiscoveryPeriod`: `checkArgument(interval >= 0)` | 0 is legal, and `Math.max(1, …)` clamps it to a 1 s sweep at runtime |

So the lower bound you already have here is right and should stay strict — don't relax it to `>= 0` to match #217. The two properties share the *upper* bound and the whole-seconds granularity rule, not the floor.

**2. Why the upper bound cannot be conditional.** The reason is worth stating in the comment, because it is not obvious from the property's behaviour: the builder's precondition runs at **build time** for every consumer, whereas the value is only *read* at runtime under the condition the property documents. An `int` overflow is the only way a non-negative NiFi duration reaches the builder negative (`TIME_PERIOD_VALIDATOR` guarantees non-negative), which is what makes the bound a correctness requirement rather than a sanity limit. On #217 that distinction is load-bearing — its validation is gated on `TOPICS_PATTERN`, so an unbounded value slips past it. Here it does not bite, because this block is already unconditional.

**3. Nothing to change about gating on this PR.** Flagging that explicitly so the #217 discussion does not send you looking for a problem that is not here: `customValidate` runs this check for every consumer, and line 699 is reached regardless of `Auto update partitions`, so validation and the builder call already agree. The only gating question on this PR is the *warning* at line 548, which is a log-noise issue, not a correctness one.

Net effect: both properties end up with "whole seconds, within `int` range", differing only in whether zero is allowed — and each floor is what its own client setter actually enforces, rather than one rule imposed on both.

jselvareef:
Agreed on the approach, and taken in `b2b2d70`. The representability framing is the right one: with the unit always in the value, the only question per property is whether the client can hold what was typed, and that decides all three rules without a judgement call. PR body revised.

**Whole seconds, 1 to `Integer.MAX_VALUE`, for *Auto Update Partition Interval*.** Exactly the condition you wrote: `< 1000 || % 1000 != 0 || toSeconds > Integer.MAX_VALUE`, one message naming the property, the value and the range. Thanks for the overflow — I had checked the floor and not the ceiling, and `10000 weeks` silently running as 55 years is the same silent-wrong-value class the PR is about. Tests: `90500 millis`, `30000 days`, `10000 weeks` and `Integer.MAX_VALUE + 1` seconds rejected; `Integer.MAX_VALUE` seconds valid and applied; all four new rejections fail against the previous commit (output in the body).

**Whole milliseconds for *Expire Time of Incomplete Chunked Message*.** Agreed the trap had moved a unit down. `nanos % 1_000_000 != 0` is rejected, so `0.5 millis` and `1500000 nanos` are invalid; `0` passes and is applied as 0, and the property description now says what it means — "disables the expiry; incomplete chunks are then kept until the pending-chunk queue evicts them" — so it is a value someone chooses rather than one they arrive at.

**The `@OnScheduled` warning — removed rather than kept, with a reason.** Under the new rule a fractional value is invalid, and NiFi never schedules an invalid processor, so `@OnScheduled` cannot fire for it: the warning would have been dead code, gated or not. The "moment it goes invalid" you describe is the validation result itself, which is where the explanation now lives (`90500 millis is not one, and would be applied as a different interval or refused by the client`). If you would still rather have a line in the log at startup for an existing `flow.xml.gz`, the only place that can run is the framework's own "processor is invalid" logging, which already quotes the validation message.

**README.** The callout is scoped *since `2.1.0`* — you are right that the truncation is in every release, and the person it addresses is the one upgrading from `2.10.0.1`. The opening sentence now states the rule and the two granularities (one property in whole seconds and constrained accordingly, the other in milliseconds and applied exactly) instead of describing the pre-fix state.

**Producer side.** Filed as #230, with the detail that matters: `getPulsarProducerConfiguration` also does `.intValue()`, so the `int` wrap exists there too, and since `loadConf()` lands on the unchecked single-argument setter, `30000 days` becomes a *negative* interval that nothing refuses — worse than the consumer's loud failure. Same rule proposed there.

Verification on the head: 381/381 unit; `ConsumePulsarIT` 4/4 and `ConsumePulsarTopicSchemaIT` 6/6 against a real broker.

jselvareef:
Saw your follow-up after posting — `b2b2d70` already matches it, and `5e6eb5f` adds the one thing it lacked.

- **Floor stays strict.** The rule is `< 1000 || % 1000 != 0 || toSeconds > Integer.MAX_VALUE`, so zero is rejected here as the consumer's setter rejects it (`> 0`); I have not relaxed it to `>= 0` to match #217, and the comment now says why the two floors differ.
- **Upper bound unconditional, and the comment says why.** Your build-time-versus-runtime point is now in the code: the precondition runs when the consumer is built, for every consumer, while the value is only read under *Auto Update Partitions*; `TIME_PERIOD_VALIDATOR` guarantees a non-negative duration, so the `int` overflow is the only way the builder can see a negative — a correctness requirement, not a sanity limit.
- **Gating.** Nothing to do on the check, as you say; the warning that raised the question is gone (a rejected value never reaches `@OnScheduled`).

Comment-only commit; 381/381 unchanged.

jselvareef:
One thing to know, unrelated to the change: the *Build & Test* run for `5e6eb5f` — a comment-only commit — failed on `ConsumePulsarAckLeakTest.exclusiveSubscriptionAcksAreNotRetained`:

```
Acknowledgement Futures are being retained: 26 left after 40 triggers. The leak produced one per
acknowledgement; anything near the trigger count is that leak, not acks in flight (see issue #53)
```

The same code passed that job an hour earlier on `b2b2d70`, and the class is green on every local run, so this is the scheduling sensitivity the test's own javadoc describes (it records an earlier CI failure with 8 against a hard 3; the allowance is now `max(4, triggers / 4)` = 10, and this run saw 26). I've re-run the failed job. The invariant test in the same class — `retainedAcksDoNotGrowWithTriggerCount` — is the one that separates a leak from acks in flight; if the two fixed-allowance cases keep tripping on loaded runners, dropping their threshold in favour of that invariant might be the cleaner fix. Happy to file it as an issue if you want it tracked.

david-streamlio:
Yes — please file it.

A test that fails on a **comment-only commit** and passes on re-run of the same tree is worse than a test that fails honestly: it teaches everyone to re-run rather than read, and the next real regression gets re-run too. That is worth an issue on its own, separate from this PR.

Your read of the fix looks right to me. `retainedAcksDoNotGrowWithTriggerCount` is the assertion that actually distinguishes a leak from acks in flight, because it measures the *shape* of the growth rather than a point value on a loaded runner. The two fixed-allowance cases are measuring scheduling luck as much as retention — the history in the javadoc makes that case by itself: a hard 3, then `max(4, triggers / 4)` = 10, and this run still saw 26. Raising the allowance again just moves the threshold ahead of the noise until the next loaded runner catches up, and each raise makes the test less able to detect the leak it exists for.

Worth capturing in the issue:

- the run that failed here (26 against 10 on `5e6eb5f`, green on `b2b2d70` an hour earlier with the same test code), since it is the cleanest evidence that the input is the runner and not the code;
- the escalation history from the javadoc, which shows the threshold has been chased twice already;
- issue #53, which the assertion message already points at, so the leak this guards against stays documented;
- and the proposal itself — keep the invariant as the gate, and either drop the fixed-allowance assertions or keep them as non-gating diagnostics.

Nothing here blocks this PR. #228 is green and the change is unrelated.

jselvareef:
Filed as #233, with the four things you listed: the three runs side by side (green on `b2b2d70`, red then green on the comment-only `5e6eb5f`), the threshold history from `c688d2f` (a hard 3, then `max(4, triggers / 4)` = 10, now 26 observed), the #53 link, and the proposal — `retainedAcksDoNotGrowWithTriggerCount` as the gate, the two fixed-allowance cases dropped or kept as non-gating diagnostics. One addition: the invariant test runs `Shared` only, so if the per-subscription cases go, it could take the subscription type as a parameter to keep `Exclusive` exercised.

And thanks for folding the release notes into this branch — pulled your merge into my worktree as-is, nothing to redo.

PR #229, opened 2026-09-08
fix: acknowledge every message individually, never cumulatively (#223)

### LINKED ISSUE #223: Cumulative acknowledgement loses messages on a non-Shared subscription with Concurrent Tasks > 1
On a non-Shared subscription (`Exclusive` or `Failover`) with **Concurrent Tasks > 1**, one task's
cumulative acknowledgement can acknowledge messages that another task is still holding — and if that
other task then fails to write them, they are neither in NiFi nor recoverable from Pulsar.

This is pre-existing, not introduced by #216 or #217. It was found while reviewing a test added in
#217 that ran `ConsumePulsar` at `setThreadCount(2)` on an `Exclusive` subscription.

### Why concurrent tasks share one consumer

`AbstractPulsarConsumerProcessor.getConsumer()` is `synchronized` and caches by consumer id
(topics + subscription name + consumer name), so every concurrent task of one processor receives from
the **same** `Consumer` instance. That part is deliberate and correct — it is what stops a second task
being refused by the broker with "Exclusive consumer is already connected".

### Why sharing it is unsafe for acknowledgement

`acknowledge()` branches on the subscription type:

```java
if (shared) {
    for (final Message<GenericRecord> message : messages) { consumer.acknowledge(message); }
} else {
    final Message<GenericRecord> last = messages.get(messages.size() - 1);
    consumer.acknowledgeCumulative(last);
}
```

Cumulative acknowledgement acknowledges the given message **and everything before it on the
subscription**. That is safe when one task owns the whole stream. It is not safe when two tasks are
draining the same consumer concurrently, because "everything before it" includes messages that belong
to the other task's in-flight batch.

### The losing interleaving

1. Task A receives `msg1`. Task B receives `msg2` from the same consumer.
2. Task B writes its FlowFile, commits, and calls `acknowledgeCumulative(msg2)` — which acknowledges
   `msg1` as well.
3. Task A's write fails (full content repository, disk fault) and it calls `rollbackAndRedeliver`.
4. `msg1` is already acknowledged, so the negative acknowledgement has nothing to redeliver. The
   broker never sends it again.

`msg1` is gone: never in NiFi, not recoverable from Pulsar. This is the same class of loss #216's
commit-then-acknowledge ordering was written to eliminate, reached by a different route.

The negative acknowledgement added in #216 makes step 4 quieter, not worse: before it, the message
would also not be redelivered, because it was already acked.

### Scope

- Needs a **non-Shared** subscription (`Exclusive` / `Failover`) and **Concurrent Tasks > 1**.
- `Shared` and `Key_Shared` are unaffected: they acknowledge per message.
- It is reachable by default — nothing warns, and `Concurrent Tasks` is a standard NiFi control.
- **`Read Compacted` (#217) makes it more likely to be hit**, because it *requires* `Exclusive` or
  `Failover`, so anyone using it is on the affected path by construction.

### Options

- Acknowledge per message on every subscription type. Simplest and always correct; loses the
  efficiency of one cumulative ack per batch, which is the reason the branch exists.
- Keep cumulative acknowledgement but make the receive-plus-acknowledge span exclusive per consumer,
  so two tasks cannot hold overlapping batches from one consumer.
- Reject `Concurrent Tasks > 1` on a non-Shared subscription at validation, and document `Shared` as
  the way to consume with more than one task. Cheapest, and consistent with how the bundle already
  rejects configurations the client cannot honour — but it removes a capability that works today for
  anyone whose writes never fail.

Whichever way it goes, the interaction deserves a note in the consumer documentation: today nothing
connects *Concurrent Tasks* to the subscription type.

Refs #195, #216, #217.

david-streamlio:
Closes #223.

Concurrent tasks of one processor share a consumer — `getConsumer()` is synchronized and caches by
consumer id, deliberately, so the broker never sees a second consumer on an `Exclusive` subscription.
Cumulative acknowledgement is safe only while one task owns the whole stream, and that was never true
here: `acknowledgeCumulative(last)` means *this message and everything before it on the subscription*,
and everything before it included whatever another task was still holding.

**The losing interleaving.** Task B commits its batch and acknowledges cumulatively, which
acknowledges task A's older message too. Task A then fails to write and negatively acknowledges —
which has nothing left to redeliver. The message is neither in NiFi nor recoverable from Pulsar: the
same loss the commit-then-acknowledge ordering was introduced to eliminate, reached by another route.

### The fix removes a distinction rather than adding a guard

Every subscription type now acknowledges per message. **This is not a new code path** — it is the loop
`Shared` subscriptions have always run, now taken by all of them. `shared` turned out to be used for
nothing else, so it goes: from `commitAndAcknowledge`, from `acknowledge`, and from the locals in both
consumers.

Deliberately **not** `acknowledge(List<MessageId>)`, which would be one call rather than a loop. This
bundle has never used that overload, and introducing an unexercised client API inside a data-loss fix
is the wrong trade — particularly to save a loop the client coalesces anyway: individual
acknowledgements are grouped into a single command over a 100ms window, up to 1000 ids.

### Fails first, both levels

**Unit** — `ConsumePulsarConcurrentAckTest`. Restoring cumulative acknowledgement fails 2 of its 3
cases. The `Shared` case passes either way, which is the point: that path was never affected.

**Integration** — `ConsumePulsarConcurrentAckIT`, real broker. It asserts on the subscription's
**backlog** after the second trigger commits, rather than waiting for a redelivery, so it never has to
distinguish "lost" from "not yet". Restoring cumulative acknowledgement fails it in the same twelve
seconds the passing run takes:

```
java.lang.AssertionError: the message rolled back by the first trigger was acknowledged by the
second trigger's commit: the subscription has no backlog, so nothing will ever redeliver it
```

### Two earlier versions of that IT were wrong

Both passed while proving nothing, and the comments in the test record why, because the setup is what
makes or breaks it:

- **A one-second redelivery delay** meant the rolled-back message came back *before* the next was
  acknowledged — the interleaving the defect needs never occurred. Caught by mutation testing.
- **Scheduling against a topic that already had messages**: `runner.run(1, false, true)` also triggers,
  so it consumed `first` and shifted the whole sequence by one. That version *did* fail against the
  bug — by timeout, with the real assertion never reached. Caught by reading why it failed rather than
  that it failed.

### Verified against `main`, not just against a mutation

Both tests were replayed onto unmodified `main` (`35d9212` — the commit `v2.11.0.1` was cut from). The
only edit was the `commitAndAcknowledge` signature, which gains a `shared` parameter there; no
assertion or setup was touched.

| | on `main` | with this fix |
|---|---|---|
| `ConsumePulsarConcurrentAckTest` | **2 of 3 fail** | 3/3 pass |
| `ConsumePulsarConcurrentAckIT` | **fails**, 11.65s | passes, 12.01s |

The IT fails on its assertion, not by timeout, and the `Shared` unit case passes on `main` — that path
genuinely was never affected, so the tests are not failing indiscriminately.

**This means the loss is reachable on the current release.** A flow on `2.11.0.1` with an `Exclusive`
or `Failover` subscription loses any message whose write fails while a later message is committed.

### Existing tests

The non-`Shared` assertions checked for one cumulative call and now check one acknowledgement per
message — **stronger, not weaker**. The 2.10.0 commit-before-acknowledge invariant is untouched and
still asserted throughout.

### Verification

370 unit, 75 integration, all green.

### Note for whoever merges second

#217 currently documents a workaround for this defect — *"Run these flows with a single task until
#223 is resolved"* — in the README and both consumer pages, because `Read Compacted` forces
`Exclusive`/`Failover`. That guidance is wrong once this lands, and needs reconciling in whichever
branch merges second.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,10 @@ decides whether Pulsar ever delivers it again.
 | The Pulsar client itself failed | rolled back, left unacknowledged | Yes, after *Acknowledgment Timeout* |
 
 Acknowledgement happens only after the FlowFile carrying the message is committed, so a message
-is never acknowledged while its content could still be discarded.
+is never acknowledged while its content could still be discarded. Each message is acknowledged
+individually, on every subscription type, so one trigger's acknowledgement never reaches a message
+another trigger is still holding — which matters because concurrent tasks of one processor share a
+single consumer.
 
 The third row is the one to know about. When the processor cannot write a message into a FlowFile
 — a full content repository, a permissions problem, a disk fault — it rolls the session back and
--- a/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
+++ b/nifi-pulsar-processors/src/main/java/org/apache/nifi/processors/pulsar/AbstractPulsarConsumerProcessor.java
@@ -759,11 +759,10 @@ public abstract class AbstractPulsarConsumerProcessor<T> extends AbstractProcess
      * @param session  the session holding the FlowFiles the messages were written to
      * @param consumer the consumer the messages were received from
      * @param messages the messages carried by the FlowFiles in the session; cleared on return
-     * @param shared   whether the subscription is Shared or Key_Shared
      * @param async    whether to acknowledge through the asynchronous acknowledgement service
      */
     protected void commitAndAcknowledge(final ProcessSession session, final Consumer<GenericRecord> consumer,
-                                        final List<Message<GenericRecord>> messages, final boolean shared, final boolean async) {
+                                        final List<Message<GenericRecord>> messages, final boolean async) {
         if (messages.isEmpty()) {
             return;
         }
@@ -771,7 +770,7 @@ public abstract class AbstractPulsarConsumerProcessor<T> extends AbstractProcess
         final List<Message<GenericRecord>> committed = new ArrayList<>(messages);
         messages.clear();
 
-        session.commitAsync(() -> acknowledge(consumer, committed, shared, async));
+        session.commitAsync(() -> acknowledge(consumer, committed, async));
     }
 
     /**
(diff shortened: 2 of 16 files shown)


david-streamlio:
Closing as stale. Issue #223 stays open; the fix will land via a separate PR.

PR #238, opened 2026-09-09
test: measure the ack-leak after the acknowledgements complete, not while they run (#233)

### LINKED ISSUE #233: ConsumePulsarAckLeakTest's fixed in-flight allowance fails on a loaded CI runner with no code change
`ConsumePulsarAckLeakTest.exclusiveSubscriptionAcksAreNotRetained` failed in CI on a commit that changed only a code comment, and passed on re-run of the same tree. A test that does that teaches everyone to re-run rather than read, and the next real regression gets re-run too.

### The run

| Run | Head | Change in the commit | Result |
|---|---|---|---|
| [34296899298](https://github.com/david-streamlio/pulsar-nifi-bundle/actions/runs/34296899298) | `b2b2d70` | validation rules (#228) | green |
| [34298571820](https://github.com/david-streamlio/pulsar-nifi-bundle/actions/runs/34298571820), attempt 1 | `5e6eb5f` | **a code comment**, nothing else | **red** |
| same run, attempt 2 | `5e6eb5f` | re-run, same tree | green |

```
ConsumePulsarAckLeakTest.exclusiveSubscriptionAcksAreNotRetained:129
Acknowledgement Futures are being retained: 26 left after 40 triggers. The leak produced one per
acknowledgement; anything near the trigger count is that leak, not acks in flight (see issue #53)
```

The other 379 tests passed on the red attempt. Locally the class is green on every run. The input is the runner's load, not the code.

### What the test measures, and why the threshold keeps moving

The class guards against #53: in async mode the acknowledgements go to an `ExecutorCompletionService`, whose completed `Future`s nobody took, so the queue grew by one per acknowledgement for the life of the processor. The two per-subscription cases (`sharedSubscriptionAcksAreNotRetained`, `exclusiveSubscriptionAcksAreNotRetained`) run 40 triggers and assert that the completion service holds at most `inFlightAllowance(40)` = `max(4, 40 / 4)` = 10 `Future`s afterwards.

That number is a guess at how many acknowledgements can legitimately still be queued when the last trigger returns, and the test's own history says it cannot be pinned:

- the first version asserted a hard **3**, on the reasoning that the ack pool has `MAX_ASYNC_REQUESTS + 1` threads; that reasoning was wrong — the pool's work queue is unbounded — and it failed in CI with **8** while passing on a developer machine (`c688d2f`, #164);
- #164 replaced it with `max(4, triggers / 4)` = **10**, "keeping a 4x margin below the leak while tolerating scheduling noise";
- this run saw **26**.

Raising the allowance again moves the threshold ahead of the noise until the next loaded runner catches up, and every raise brings it closer to the 40 the leak itself would produce, so the assertion detects less of the thing it exists for each time.

### The assertion that does not depend on the runner

#164 also added `retainedAcksDoNotGrowWithTriggerCount`, which measures the *shape* of the growth rather than a point value: 40 triggers and 160 triggers, asserting that the larger run does not retain proportionally more. With the leak the two numbers were 40 and 160; without it they stay in one range however far apart the trigger counts are. That is the property that separates a leak from acks in flight, and it survives a loaded runner because both measurements are taken under the same load.

### Proposal

Keep `retainedAcksDoNotGrowWithTriggerCount` as the gate. For the two fixed-allowance cases, either drop the threshold assertion or keep them as non-gating diagnostics — log the retained count, do not fail on it — so a loaded runner cannot turn a comment change red. The one thing worth preserving from them is that both subscription types are exercised; the invariant test runs `Shared` only today, and could take the subscription type as a parameter.

Refs #53, #164, #228.

jselvareef:
Closes #233.

`ConsumePulsarAckLeakTest` bounded how many acknowledgement `Future`s may still be queued at the instant the last trigger returns. That number is how far the ack pool lags behind the trigger thread — a property of the machine, not of the processor — and the history in #233 shows it cannot be pinned: a hard 3 failed in CI with 8 (#164), a quarter of the triggers failed with 26 on a comment-only commit and with 21 on a Dependabot bundle, in one of two concurrent jobs on the same tree.

### What changed

The leak is a property of what happens **after** an acknowledgement completes: with the drain in place the next trigger takes the completed `Future`; without it the `Future` stays queued for good. So the gate now measures exactly that, and nothing else:

1. run the triggers;
2. wait until every acknowledgement they submitted has completed (`ThreadPoolExecutor.getTaskCount() == getCompletedTaskCount()`, 30 s ceiling);
3. run one more trigger on the now-empty topic — it receives nothing and only drains;
4. assert that **zero** `Future`s are left.

No allowance to tune. With the drain removed the assertion fails with exactly 40 retained after 40 triggers, for both subscription types.

- **Parameterised by subscription type** (`Shared`, `Exclusive`), so the cumulative-ack path keeps its coverage — your point on #233 that the old invariant test ran `Shared` only.
- **The in-flight count at the moment the last trigger returns is printed, not asserted** — the diagnostic you suggested keeping, so a loaded runner still shows in the log (`Exclusive: 1 acknowledgement(s) still pending when the last of 40 triggers returned` here).
- **Removed:** `sharedSubscriptionAcksAreNotRetained` and `exclusiveSubscriptionAcksAreNotRetained` (the flaky assertions), and `retainedAcksDoNotGrowWithTriggerCount`. I went one step past the proposal on the last one: its first assertion was itself a fixed allowance (`afterMany <= 40` at 160 triggers), and its shape assertion tolerated 30 extra `Future`s at four times the triggers — with 26 observed at 40 triggers on a loaded runner, that margin is of the same kind. Once the measurement waits for completion the shape comparison has nothing left to distinguish, so the deterministic gate replaces it rather than sitting beside it.
- **Kept:** `idleTopicDoesNotQueueFailedAcks`, with the same wait before counting.

### Fails first

With `drainAcknowledgments()` removed from `ConsumePulsar`'s `finally`:

```
Tests run: 4, Failures: 2, Errors: 0 -- in ConsumePulsarAckLeakTest
  completedAcknowledgementsAreDrainedByTheNextTrigger[Shared]:    expected:<0> but was:<40>
  completedAcknowledgementsAreDrainedByTheNextTrigger[Exclusive]: expected:<0> but was:<40>
```

### Verification

- `mvn test`: **426/426**.
- The new class, 5 consecutive runs with 40 busy-loop processes on a 20-core machine (the class took 16–22 s instead of 4 s): 4/4 every time.
- What I could **not** do: reproduce the old test's CI failure locally. Under the same load the previous version passed 3/3 here. A 20-core box does not starve the ack threads the way a 2–4 vCPU runner does, so the fix is argued from the measurement's construction (it no longer depends on lag at all) and the mutation, not from a local red of the old assertion.

### Not done

- `ConsumePulsarRecord` has the same drain in its `finally` and no test of its own for the leak; this class tests `ConsumePulsar` only, as before.
- The 30 s ceiling on the completion wait is a guard against a hung pool, not a tuning knob; on every run here the wait was under a second.

https://claude.ai/code/session_016FxYKxcaKpvRbJSYP3QYZ9

### DIFF
--- a/nifi-pulsar-processors/src/test/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarAckLeakTest.java
+++ b/nifi-pulsar-processors/src/test/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarAckLeakTest.java
@@ -19,6 +19,7 @@ package org.apache.nifi.processors.pulsar.pubsub;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeTrue;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -85,7 +86,12 @@ public class ConsumePulsarAckLeakTest extends AbstractPulsarProcessorTest<Generi
         long pendingAcks() {
             final ExecutorService pool = getAckPool();
             if (!(pool instanceof ThreadPoolExecutor)) {
-                return 0;
+                // Loud on purpose. Returning 0 here would make awaitSubmittedAcksToComplete() return without
+                // waiting and pass its own check trivially, and the gate would silently go back to measuring
+                // acks in flight on a loaded machine - the flake this test exists to be rid of (#233).
+                throw new AssertionError("the ack pool is a " + (pool == null ? "null" : pool.getClass().getName())
+                        + "; pendingAcks() can only measure a ThreadPoolExecutor, so this test cannot take the "
+                        + "machine out of the measurement (see #233)");
             }
             final ThreadPoolExecutor executor = (ThreadPoolExecutor) pool;
             return executor.getTaskCount() - executor.getCompletedTaskCount();
@@ -166,9 +172,15 @@ public class ConsumePulsarAckLeakTest extends AbstractPulsarProcessorTest<Generi
      * The aggravating case: an idle topic. The cumulative-ack task used to be submitted outside the
      * "did we receive anything?" guard, so every trigger queued a Future holding an
      * IndexOutOfBoundsException from messages.get(-1) - an idle processor leaked fastest of all.
+     * <p>
+     * That is a defect of the cumulative path, which only Exclusive takes: a Shared subscription acknowledges
+     * per message and submits nothing on an idle topic, so its run of this test could only ever assert
+     * {@code 0 == 0}. Pinned to Exclusive; the gate test above is where both types earn their place.
      */
     @Test
     public void idleTopicDoesNotQueueFailedAcks() throws Exception {
+        assumeTrue("the idle-topic defect lives on the cumulative-ack path, which only Exclusive takes",
+                "Exclusive".equals(subscriptionType));
         mockClientService.setMockMessageQueue(new ArrayList<>());
 
         runner.run(TRIGGERS, false);
--- a/nifi-pulsar-processors/src/test/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarAckLeakTest.java
+++ b/nifi-pulsar-processors/src/test/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarAckLeakTest.java
@@ -17,11 +17,16 @@
 package org.apache.nifi.processors.pulsar.pubsub;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.nifi.processors.pulsar.AbstractPulsarConsumerProcessor;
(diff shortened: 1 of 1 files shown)


nifi-pulsar-processors/src/test/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarAckLeakTest.java
```diff
@@ -33,50 +38,80 @@
...
+
+        /**
+         * How many submitted acknowledgements have not run yet. This is the number the old fixed allowance
+         * tried to bound, and it is a property of the machine, not of the processor: logged for the record,
+         * never asserted on.
+         */
+        long pendingAcks() {
+            final ExecutorService pool = getAckPool();
+            if (!(pool instanceof ThreadPoolExecutor)) {
+                return 0;
```
david-streamlio: **This fallback can silently reintroduce the exact flake this PR removes.**

`Executors.newFixedThreadPool` returns a `ThreadPoolExecutor` today, so the `instanceof` holds and everything works. But consider what happens if the ack pool is ever changed — a virtual-thread executor, or anything wrapped:

- `pendingAcks()` returns `0`;
- `awaitSubmittedAcksToComplete()` sees `0 > 0` as false and returns immediately, without waiting for anything;
- its own guard, `assertEquals("acknowledgements still running after 30 s", 0L, pendingAcks())`, passes trivially because both sides are the constant `0`.

So the wait disappears with no signal at all, and the gate silently reverts to measuring acks in flight on a loaded runner — which is precisely the failure mode #233 is about. The test would go back to being red on comment-only commits, and the reason would be invisible.

Since the wait is the mechanism that takes the machine out of the measurement, its precondition deserves to fail loudly rather than degrade quietly:

```java
if (!(pool instanceof ThreadPoolExecutor)) {
    throw new AssertionError("the ack pool is a " + (pool == null ? "null" : pool.getClass().getName())
            + "; pendingAcks() can only measure a ThreadPoolExecutor, so this test cannot take the "
            + "machine out of the measurement (see #233)");
}
```

That way a future pool change breaks the test with an explanation instead of making it flaky again.

nifi-pulsar-processors/src/test/java/org/apache/nifi/processors/pulsar/pubsub/ConsumePulsarAckLeakTest.java
```diff
@@ -184,10 +169,10 @@ private int retainedAfter(final int triggers) throws Exception {
      */
     @Test
     public void idleTopicDoesNotQueueFailedAcks() throws Exception {
-        runner.setProperty(AbstractPulsarConsumerProcessor.SUBSCRIPTION_TYPE, "Exclusive");
         mockClientService.setMockMessageQueue(new ArrayList<>());
 
         runner.run(TRIGGERS, false);
+        processor.awaitSubmittedAcksToComplete();
```
david-streamlio, line 187: **Minor — half of this test's parameterisation is now vacuous, and its javadoc describes only the other half.**

This test previously forced `Exclusive`, and the javadoc above says why: the aggravating case is the **cumulative**-ack task being submitted outside the "did we receive anything?" guard, queueing a Future that holds an `IndexOutOfBoundsException` from `messages.get(-1)`. That is an `Exclusive`/`Failover` defect — `Shared` acknowledges per message and never takes that path.

Now that the class is parameterised, the `Shared` run of this test drives an idle topic that produces no acknowledgements at all, then asserts none were retained: `0 == 0`, true for any implementation. Harmless, and the `Exclusive` run still carries the real regression value — but the javadoc now names a defect only one of the two runs can detect.

Either note that in the comment, or keep this one pinned to the type whose defect it describes (the gate test is where both types genuinely earn their place).

david-streamlio (approved):
Approving. This is a **better fix than the one I signed off on in #233**, and worth saying why: that issue proposed keeping `retainedAcksDoNotGrowWithTriggerCount` as the gate and demoting the fixed-allowance cases to diagnostics — tolerating the machine-dependent variable. You removed it instead. Waiting for the acks to *complete* and then asserting exactly zero is the stronger move, and it makes the allowance question disappear rather than relocating it.

The assertion went from `retained <= 10` to `retained == 0`.

## Verified against the code rather than the description

- **`drainAcknowledgments()` really does drain fully** — `while (ack != null)` over a non-blocking `poll()`, so one trigger clears the whole queue rather than a bounded slice.
- **It really does run on an empty trigger** — it sits in a `finally` in `ConsumePulsar.onTrigger` (line 205), so "this trigger receives nothing and only drains" holds.
- **There is no completion race**, which was my main worry about `getTaskCount() - getCompletedTaskCount()`. Traced the ordering: `FutureTask.run()` → `finishCompletion()` → `done()` enqueues into the `ExecutorCompletionService` *before* `ThreadPoolExecutor` increments `completedTaskCount`. So `pendingAcks() == 0` guarantees every Future is already queued, and the wait cannot finish early. That ordering is what makes the whole approach work, and it is load-independent.
- **Bidirectional validity.** I removed the `drainAcknowledgments()` call to restore the #53 leak: both parameterised variants failed with **exactly 40 retained against expected 0**. The signal is now 40-vs-0 where it was 21-vs-10 — a margin no runner load can close.
- **Stability.** Five consecutive local runs, 4/4 each, 3.15–3.19 s: a 0.03 s spread. Full unit suite 426/426.

Parameterising by subscription type is a real gain rather than tidying: `Shared` acknowledges individually and `Exclusive` cumulatively, and both now go through the gate instead of one of them.

## Two judgement calls I agree with, noted for the record

**Deleting `retainedAcksDoNotGrowWithTriggerCount`** is right. With in-flight acks out of the measurement, a scaling test has nothing left to distinguish — its whole purpose was separating "a few in flight" from "a leak". The only thing lost is the 160-trigger long run, which was never its stated point. Flagging it only because it diverges from what #233 proposed, so the issue's plan and this PR should not be read as the same thing.

**`ConsumePulsarRecord` still has no coverage** for its own `drainAcknowledgments()` call (`ConsumePulsarRecord.java:758`) — `AckProbeConsumePulsar extends ConsumePulsar`, so the record processor's independent copy of that call is untested. Pre-existing and not a regression, but this PR is specifically about making the leak test trustworthy, and that is the remaining gap in it. Reasonable as a follow-up.

Two inline findings, neither blocking. The first is the one I would fix before merge — it is small, and left as is it can silently undo this entire PR.

david-streamlio:
Approved above. Two small things I'd like before this merges, and then the third item from my review is filed rather than blocking.

**1. The `instanceof ThreadPoolExecutor` fallback** (inline thread on line 88). This is the only one I'd hold the merge for, and the reason is asymmetric with everything else: deferring it leaves a trap that disables the very assertion this PR adds. If the ack pool ever becomes a virtual-thread or wrapped executor, `pendingAcks()` returns `0`, `awaitSubmittedAcksToComplete()` returns without waiting, and its own `assertEquals(0L, pendingAcks())` passes because both sides are the constant — so the gate quietly goes back to measuring acks in flight, and the flakiness returns with nothing pointing at the cause. Throwing there instead makes a future pool change break the test with an explanation. Four lines.

**2. The javadoc on `idleTopicDoesNotQueueFailedAcks`** (inline thread on line 175). Only because the fix is cheaper than tracking it: the aggravating case that comment describes is the cumulative-ack path, which `Shared` never takes, so half the parameterisation asserts `0 == 0`. A sentence noting that, or pinning this one test back to `Exclusive`, either is fine — the gate test is where both types genuinely earn their place.

**3. `ConsumePulsarRecord` coverage — filed as #239**, not for this PR. It needs the probe methods to move somewhere both processors can inherit them, which is more than a tweak, and it is pre-existing rather than anything #238 introduced: the test had the same single-processor scope before, and this PR only changed how the measurement is taken.

Worth mentioning why I'd rather not grow the scope beyond those two: this flake is already costing other work. It turned #237 red on a bundle that changed six version properties in `pom.xml`, and the re-run to unblock it was exactly the habit #233 warns about. The sooner this lands, the sooner that stops.

Verification from my end, for the record: with `drainAcknowledgments()` removed both parameterised variants fail at exactly 40 retained against expected 0, and five consecutive local runs pass 4/4 within a 0.03 s spread. The 40-vs-0 margin is the part that makes this hold up on a loaded runner where 21-vs-10 did not.

jselvareef:
Both in `5939add`:

1. **Loud precondition.** `pendingAcks()` throws an `AssertionError` naming the pool's class when the ack pool is not a `ThreadPoolExecutor`, with the reason in the message — the wait is the mechanism that takes the machine out of the measurement, so losing it silently would hand the flake back with nothing pointing at the cause. Your four lines, plus a comment saying why.
2. **Idle-topic case pinned to `Exclusive`.** An `assumeTrue` on the subscription type, so the `Shared` variant shows as skipped with the reason rather than asserting `0 == 0`; the javadoc now says the defect it guards lives on the cumulative path. Class: 4 run, 1 skipped, 0 failed.

Thanks for tracing the `done()`-before-`completedTaskCount` ordering — I had relied on it without writing it down, and it is the fact the whole approach rests on. #239 noted; happy to take it as the follow-up.

